PR #38, opened 2017-11-02
Lock tracking to catch locks held too long

charmster:
Track locks to catch cases when they're held too long and log a warning if this occurs.  This is a "drop in" replacement for sync.Mutex and sync.RWMutex.  It is also used to track DLM locks, but in that case the DLM lock code had to be modified.

When a lock is acquired we record the stack trace of the caller.  When the lock is released we check how long it was held and if held too long, we log a warning that includes the stack trace when the lock was acquired and when it was released.

There is also a trackedlock watcher that periodically examines all locks to see if they have been held too long and, if so, it logs a warning for the lock that has been held the longest (along with stack trace of the goroutine that acquired the lock).

Either functionality can be disabled or enabled "dynamically". It is disabled by default.

Note: github seems to be ordering these commits by the (nominal) time they were committed instead of in the order they were committed!  the times seemed to have been messed up by my use of `git rebase -i` to tidy up my work. At this point only
```
8893283 Changed trackedlock to use sync.pool for stacktrace storage.
```
is in the wrong order; it should come right after `69452f1`, not two commits later.

Almost all of trackedlock code is in commit `69452f1`.  The earlier commits lay the groundwork for testing it (the logger changes) and make a minor bug fix.  The later commits change the trackedlock to use a sync.Pool to hold space for stack traceback (`8893283`) and subsequent commits propagate the use of trackedlock across the code base (`5be0e38`, `712826a`) and then convert DLM to use sync.Pool for its internal locks (`16b45ad`) and swiftclient to use sync.Pool for ChunkedPutContextStruct (`12d6345`). Depending on your interest you may want to focus on only some of them.

Note: The SwiftClient changes to use sync.Pool required that i add a method to sortedmap.LLRBtree to Reset() a tree to the empty state.


vendor/github.com/swiftstack/sortedmap/llrb_tree.go
```diff
@@ -36,6 +36,12 @@ type llrbTreeStruct struct {
 
 // API functions (see api.go)
 
+// Return the tree to its original, empty, state.
+//
+func (tree *llrbTreeStruct) Reset() {
+	tree.root = nil
```
edmc-ss: That should do it... although simply dropping the reference to the LLRBTree itself would accomplish the same thing. I'm going to presume that you found a case where you want to re-use an LLRBTree?

vendor/github.com/swiftstack/sortedmap/llrb_tree_api.go
```diff
@@ -2,6 +2,7 @@ package sortedmap
 
 type LLRBTree interface {
 	SortedMap
+	Reset()
```
edmc-ss: We probably want to do this in the "common" part of the API as I think it is equally needed there.
That said, I believe I did the same thing only for the B+Tree previously - oops :-).

vendor/github.com/swiftstack/sortedmap/llrb_tree.go
charmster: llrbTreeStruct has an embedded Mutex. I was trying to cut down on the instances where we allocate a Mutex and then discard it (for garbage collection to clean up) because I want to cut down on the number of "active" Mutex that the trackedlock lock watcher has to keep track of (the lock watcher gives up on Mutex and RWMutex that have not been used for 20 sec (configurable), so it will continue to hold a reference to these Mutex for 20 sec).  From both the GC perspective and the lock watcher perspective, its better to reuse the Mutex via a Pool then leave them lying around for 20 sec and then letting the GC clean them up.

vendor/github.com/swiftstack/sortedmap/llrb_tree_api.go
charmster: Making this common seems fine.  i was being lazy!

charmster:
frankly, i didn't spend any time worrying about how the changes to the sortedmap package would be made (in the base repository or in the proxyfs repo).  if we agree the additional method is a good idea, then its better to make them in sortedmap repo.  i would be delighted if you pushed this change back to the sortedmap repo so Proxyfs could incorporate it automatically!

as far as Discard() is concerned,i was debating the name for this method as Reset() or Clear().  i found that some other packages had a Reset() method that returned an object to the state it was in after its initial creation, so i went with tradition and named it Reset() -- i'm sorry, i don't remember the other packages i was looking at!

for me, Discard() sounds like Free() -- an indication that you're not going to use an object any more rather then an indication you want it cleared to its initial state so you can continue to use it.  i would prefer to call such an operation Reset() or Clear() or ReInit(), but if you have a strong preference for Discard() i'm not going to spend time to argue with you.  Discard() is acceptable.

vendor/github.com/swiftstack/sortedmap/llrb_tree_api.go
edmc-ss: In retrospect, the equivalent of Reset() is actually a bit non-trivial... indeed, it is so heavyweight that the Garbage Collector would be expected to actually do a better job (i.e. I'd just say the caller should New() another one dropping the reference to the prior one)... the sync.Mutex at the top is hardly significant to it.

I rather like the distinction, now, between Reset() - meaning return the tree to the state it had just after it was created - and Discard() which actually is just a hint to cause Pass One of the Garbage Collector to know sooner that the nodes of the now unreferenced B+Tree are to be collected...

vendor/github.com/swiftstack/sortedmap/llrb_tree.go
edmc-ss: I get it... the overhead of creating/destroying an LLRB tree is likely dominated by the Mutex... assuming the tree is relatively empty.

dlm/llm.go
```diff
@@ -304,12 +333,24 @@ func (l *RWLockStruct) unlock() (err error) {
...
+	// record the release of the lock
+	track.rwMutexTrack.DLMUnlockTrack(track)
 
 	// See if any locks can be granted
 	processLocalQ(track)
 
 	track.Mutex.Unlock()
 
+	// can't return the
+	if deleted {
```
bschatz-swift, line 381: Seems like a missing comment?

dlm/llm.go
```diff
@@ -304,12 +333,24 @@ func (l *RWLockStruct) unlock() (err error) {
...
+		if track.waitReqQ.Len() != 0 || track.waiters != 0 || track.state != stale {
+			panic(fmt.Sprintf(
+				"localLockTrack object %p retrieved from pool does not have an empty waitReqQ",
```
bschatz-swift, line 384: Probably does not matter but this message should be unique and not conflict with error message at line 248 above

PR #43, opened 2017-11-20
docs updates

notmyname:
- Updated docs structure
- Mograted from markdown to restructured text
- Set up sphinx to build the docs


notmyname:
I strongly suspect things can be made better (re the md->rst conversion), but the content should all still be there.

There will be more work later to set up a more fleshed out docs tree

PR #48, opened 2017-11-22
pfs_middleware: include Content-Range, etc. on 416

smerritt:
416 responses from Swift have a bunch of headers, including
Content-Range, object metadata, and so on. This commit adds these
headers to pfs_middleware's 416 response.


tipabu (approved):
Seems to make sense.

PR #49, opened 2017-11-23
Fix large suffix-byte-range GET requests.

smerritt:
Given a 100-byte file and a request for the last 101 bytes (Range:
bytes=-101), Swift will return a 206 Partial Content response with all
100 bytes included. Using a bimodal account, one would get a 416. This
was because ProxyFS was incorrectly handling such a request and
returning an empty read plan to the middleware.

This commit adds handling for such requests; now, a read plan for the
whole file will be returned to the middleware.


tipabu (approved):
Eyeballing the go code, seems to make sense. The 206-with-full-content definitely matches Swift (odd though it is).

PR #51, opened 2017-11-29
Fix container acl caching

alistairncoles:


pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -369,27 +381,36 @@ def extract_object_metadata_from_headers(headers):
...
+                header in SPECIAL_CONTAINER_METADATA_HEADERS) and
+                (req.environ.get('swift_owner', False) or
+                 header not in SWIFT_OWNER_HEADERS)):
             meta_headers[header] = value
+        if header.startswith("X-Remove-"):
+            header = header.replace("-Remove", "", 1)
+            if ((header.startswith("X-Container-Meta-") or
+                    header in SPECIAL_CONTAINER_METADATA_HEADERS) and
+                    (req.environ.get('swift_owner', False) or
+                     header not in SWIFT_OWNER_HEADERS)):
```
tipabu, line 412: Shouldn't this include sysmeta as well?

Hmm... or does the noauth-proxy's gatekeeper drop it regardless of what we do here?

tipabu (commented):
How *do* we handle sysmeta in proxyfs-land?

tipabu:
Single-user Swift func tests before:
```
Ran 485 tests in 681.425s

FAILED (SKIP=79, errors=33, failures=74)
```

And after:
```
Ran 485 tests in 755.537s

FAILED (SKIP=79, errors=30, failures=74)
```

So, an extra 3 tests pass. Sounds like a win. Adding `X-Container-Sysmeta-*` doesn't improve the numbers any, so we probably have some other work to do to get stuff like versioning to work. Let's land and iterate.

pfs_middleware/pfs_middleware/middleware.py
alistairncoles: the x-remove-* handling in swift is (as I understand it) for clients that can't get empty header values to the proxy. sysmeta can and should always be deleted by simply setting an empty value. So the behaviour above should  mimic exactly what swift proxy-server does. (Note that non-remove sysmeta is passed).

AFAICT these metadata don't get passed as headers to the noauth proxy - if they did then there would be issues with swift_owner being False (since there is no auth to set swift_owner). In fact my first reaction to some of these bug was "how come the noauth proxy isn't filtering/transforming headers?". However I don't yet have a good understanding of what happens on the proxyfs side.

alistairncoles:
I saw

before:
```
[swift@swift-runway-001 swift]$ nosetests ./test/functional/
...
Ran 485 tests in 158.671s

FAILED (SKIP=24, errors=25, failures=107)
```
after:

```
Ran 485 tests in 189.331s

FAILED (SKIP=24, errors=21, failures=95)
```

PR #54, opened 2017-12-04
Add OPTIONS support to pfs_middleware

alistairncoles:
...by adding the ability to forward requests to a proxy server
controller that alredy implements the OPTIONS verb. For this
to work, the controller is instantiated with a fake proxy server
app that provides some config attributes and forwards any
subrequests back to the bimodal checker.


smerritt:
New plan: do https://review.openstack.org/#/c/528106/ instead

PR #55, opened 2017-12-07
Stop allowing directory-traversal in object-store paths

tipabu:
Previously, this allowed users to circumvent container ACLs; you could walk right out of the container you had access to and into one you weren't supposed to see.

Now, all paths within a bimodal account that have '.' or '..' as a component will either 400 or 404 depending on context. While we're at it, also prevent paths like

    /v1/a/c//o
    /v1/a/c/o/

since they currently resolve to

    /v1/a/c/o

which (under the Swift API) are separate things.

### DIFF
--- a/pfs_middleware/pfs_middleware/middleware.py
+++ b/pfs_middleware/pfs_middleware/middleware.py
@@ -653,13 +653,28 @@ class PfsMiddleware(object):
     def __call__(self, req):
         vrs, acc, con, obj = utils.parse_path(req.path)
 
-        if not acc or not constraints.valid_api_version(vrs):
+        if not acc or not constraints.valid_api_version(vrs) or (
+                obj and not con):
             # could be a GET /info request or something made up by some
             # other middleware; get out of the way.
             return self.app
         if not constraints.check_utf8(req.path_info):
             return swob.HTTPPreconditionFailed(
                 body='Invalid UTF8 or contains NULL')
+
+        if con in ('.', '..'):
+            if req.method == 'PUT' and not obj:
+                return swob.HTTPBadRequest(
+                    request=req, body='Container name cannot be "." or ".."')
+            else:
+                return swob.HTTPNotFound(request=req)
+        elif obj and any(p in ('', '.', '..') for p in obj.split('/')):
+            if req.method == 'PUT':
+                return swob.HTTPBadRequest(
+                    request=req,
+                    body='No path component may be "", ".", or ".."')
+            else:
+                return swob.HTTPNotFound(request=req)
 
         try:
             # Check account to see if this is a bimodal-access account or
--- a/pfs_middleware/pfs_middleware/middleware.py
+++ b/pfs_middleware/pfs_middleware/middleware.py
@@ -712,7 +712,7 @@ class PfsMiddleware(object):
                 elif method == 'DELETE' and obj:
                     resp = self.delete_object(ctx)
                 elif method == 'COALESCE' and obj:
-                    resp = self.coalesce_object(ctx)
+                    resp = self.coalesce_object(ctx, auth_cb)
 
                 elif method == 'GET' and con:
                     resp = self.get_container(ctx)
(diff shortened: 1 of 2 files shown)


FoxLisk:
this kind of seems like a big hammer. seems like allowing relative paths would be a good thing if we could just, like, do ACLs right on them. is that not possible for some technical reason I'm not aware of? with my very limited knowledge of the system, it seems like you could canonicalize the path and _then_ ask proxyfs about access rights?

tipabu:
I considered that, but in Swift objects don't follow filesystem rules. You can upload

    curl -v http://saio:8090/v1/AUTH_test/c/o -X PUT -d 1234
    curl -v http://saio:8090/v1/AUTH_test/c/%2e%2e/o -X PUT -d 5678
    curl -v http://saio:8090/v1/AUTH_test/c/%2e/o -X PUT -d abcd
    curl -v http://saio:8090/v1/AUTH_test/c//o -X PUT -d efgh

(noting that `%2e` is the url-encoded form of ".") and when you look at your listings, those are all *distinct objects*

    curl -v http://saio:8090/v1/AUTH_test/c?format=json
    ...
    [{"bytes": 4, "last_modified": "2017-12-08T02:20:42.376120",
      "hash": "674f3c2c1a8a6f90461e8a66fb5550ba", "name": "../o",
      "content_type": "application/x-www-form-urlencoded"},
     {"bytes": 4, "last_modified": "2017-12-08T02:20:51.435890",
      "hash": "e2fc714c4727ee9395f324cd2e7f331f", "name": "./o",
      "content_type": "application/x-www-form-urlencoded"},
     {"bytes": 4, "last_modified": "2017-12-08T02:21:41.601100",
      "hash": "1f7690ebdd9b4caf8fab49ca1757bf27", "name": "/o",
      "content_type": "application/x-www-form-urlencoded"},
     {"bytes": 4, "last_modified": "2017-12-08T02:20:25.058550",
      "hash": "81dc9bdb52d04dc20036dbd8313ed055", "name": "o",
      "content_type": "application/x-www-form-urlencoded"}]

I'm rather nervous about object-API users being surprised that they overwrote something they didn't intend to.

charmster:
I'm not sure how, proxyfs does checking of container based permissions.  is that done in the middleware instead?

if ACLs on containers were propagated into the file system then proxyfsd could check whether testuser:test has permission to search (or modify) /v1/AUTH_test/bar when testuser:test tried to `..` its way out of container "foo" and into "bar. 

but it seems like the bigger problem is how to handle non-canonical pathnames! as a filesystem developer i feel like the Swift names "a", "./a", "./././a", "./////a", "../foo/a", ad nauseum should all map to the same object without error, but @tipabu makes a very good point about these normally being distinct objects.
the alternative seems to be returning an error for any object name that is not a canonical pathname, so all but one of the above pathnames would get an error.  i guess that's the thing to do since it addresses both problems.
and another thing!  if we did let a client create "./././a" when they performed a listing of objects in the container they would only see "a"!

tipabu:
Yeah, container ACLs currently get checked in middleware; see https://github.com/swiftstack/ProxyFS/blob/1.0.3/pfs_middleware/pfs_middleware/middleware.py#L671-L690

They're persisted as an alternate data stream on the dir object, as I recall. Maybe we could have some plumbing to enforce those permissions on the FS side? But I have no idea how we'd pass auth info for the "current user" (or what that would even look like).

smerritt:
Conceptually, this certainly closes the hole, but it also closes off some (arguably-) legitimate requests. Still, it's easier to open things up than to close them off, so if we're going to do this, it should be sooner rather than later.

Code-wise, there's a bit more to do. How would it look to pull all the path validation up into `PfsMiddleware.__call__`? Right now, there's a bit too much duplication IMO. Also, `COALESCE` requests need filename validation, and for that, you'll want to check both the request path and all the file paths in the request body.

tipabu:
Added authorization checks to `COALESCE` requests -- the wisdom of allowing cross-container (or even more generally, cross-directory) coalescing is certainly up for debate, but I think beyond the scope of what I want to do here.

Good idea on pulling the validation up to `__call__` -- took out about half the number of lines, and makes the logic easier to follow.

tipabu:
OK, now there are similar restrictions on `elements` items -- no double slashes, no `.` or `..` components. I think this is fairly reasonable now.

smerritt:
Alright, I think this is saner than it was. I'm not 100% sure how well the element authentication stuff is going to work with ACLs, but it'll at least work for containers where you've got swift_owner privileges.

PR #56, opened 2017-12-08
Bug fixes for permission checking

charmster:
Fix sundry errors or omissions in permission checking.  The specific problem is that a tar archive failed to unpack because some of the files were read-only (0400) and tar sets the permission on the files before it writes the content.  This works on a local file system (where tar has a filehandle that is open for writing), but not over NFS.
Allow the owner of a file to write to it even the permission bits deny it.  Do the same for other requests, like Flush(), GetXAttr(), Resize() (i.e. Trunc()), Flock(), etc. (see the ones with `inode.OwnerOverride`.
Fix other issues: only the owner of the file can change the permissions on the file, eXecute (search) permission is not required to read the contents of a directory, etc.
Get rid of `inode.InodeRootGroupID` because there is no such as a "root group".
See the comments in the commits for more info.

### DIFF
--- a/fs/api_internal.go
+++ b/fs/api_internal.go
@@ -155,7 +155,8 @@ func (vS *volumeStruct) inFlightFileInodeDataFlusher(inodeNumber inode.InodeNumb
 		logger.PanicfWithError(err, "dlm.Writelock() for volume '%s' inode %d failed", vS.volumeName, inodeNumber)
 	}
 
-	stillExists = vS.VolumeHandle.Access(inodeNumber, inode.InodeRootUserID, inode.InodeGroupID(0), nil, inode.F_OK)
+	stillExists = vS.VolumeHandle.Access(inodeNumber, inode.InodeRootUserID, inode.InodeGroupID(0), nil, inode.F_OK,
+		inode.NoOverride)
 	if stillExists {
 		err = vS.VolumeHandle.Flush(inodeNumber, false)
 		if nil != err {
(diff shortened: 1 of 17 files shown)


fs/api_internal.go
```diff
@@ -1289,7 +1305,7 @@ func (mS *mountStruct) MiddlewareGetAccount(maxEntries uint64, marker string) (a
 	lastBasename := marker
 	for areMoreEntries && uint64(len(accountEnts)) < maxEntries {
 		var dirEnts []inode.DirEntry
-		dirEnts, _, areMoreEntries, err = mS.Readdir(inode.InodeRootUserID, inode.InodeRootGroupID, nil, inode.RootDirInodeNumber, lastBasename, maxEntries-uint64(len(accountEnts)), 0)
+		dirEnts, _, areMoreEntries, err = mS.Readdir(inode.InodeRootUserID, inode.InodeGroupID(0), nil, inode.RootDirInodeNumber, lastBasename, maxEntries-uint64(len(accountEnts)), 0)
```
edmc-ss, line 1394: I do not agree with the change to convert inode.InodeRootGroupID to inode.InodeGroupID(0)...
There is, in fact, a "root" GroupID being zero on both CentOS and Ubuntu from what I can see and do not think we need to "avoid" it in the calls.

That said, as far as I understand it, the fact that inode.InodeRootUserID is passed obviates the need to even ever consult the gid argument :-).

charmster: thanks for your reply.  i'm sorry i didn't get to it sooner!

the problem that i see with "InodeRootGroupID" is that people seem to assume it has special priviledges, probably because InodeRootUserID does (see the code in Access() that i deleted), but there are no special privileges associated with Group ID 0 so i don't think it deserves a special name.  for example, we don't have a InodeBinGroupID (gid 2 on Ubuntu and Centos and probably everywhere else).

so i wanted to demote it to "place holder" status.

what if we rename it to InodeGroupDummyID?

charmster: it looks like MiddlewareCoalesce() creates a directory uses InodeRootUserID and InodeRootGrooupID (UID 0 and GID 0) and permissions 0755.

however, putObjectHelper() (for MiddlewarePutComplete()) just uses 0 for the UID and GID and 0777 for permissions on the directory(s) and file.  it also bypasses permission checking on its way to the file, so i think MiddlewarePut() can overwrite any file in the file system.  the new objects probably shouldn't be executable!

edmc-ss (approved):
This looks just fine...

My initial reaction to obsoleting inode.InodeRootGroupID was because I mistakenly thought the root's group ID had a similar behavior as does the root's owner ID. I gather this is not correct nor even useful. If somebody wants to bypass the Access() check by being Root, it matters not what their egid is :-).

PR #57, opened 2017-12-08
Add tracing and Log with usec timestamps

charmster:
This branch is mis-named since it doesn't fix any bugs in the sortedmap code, though that was its genesis.
Add tracing messages to parts of the swiftclient code and the file system code that updates the object map (extent map) and flushes both inodes and B+ trees.

### DIFF
--- a/fs/api_internal.go
+++ b/fs/api_internal.go
@@ -170,13 +170,15 @@ func (vS *volumeStruct) inFlightFileInodeDataFlusher(inodeNumber inode.InodeNumb
 }
 
 func (inFlightFileInodeData *inFlightFileInodeDataStruct) inFlightFileInodeDataTracker() {
+
+	logger.Tracef("fs.inFlightFileInodeDataTracker() entry: waiting to flush volume '%s' inode %d",
+		inFlightFileInodeData.volStruct.volumeName, inFlightFileInodeData.InodeNumber)
+	defer logger.Tracef("fs.inFlightFileInodeDataTracker() return: flush volume '%s' inode %d",
+		inFlightFileInodeData.volStruct.volumeName, inFlightFileInodeData.InodeNumber)
+
 	var (
 		flushFirst bool
 	)
-
-	logger.Tracef("fs.inFlightFileInodeDataTracker(): waiting to flush volume '%s' inode %d",
-		inFlightFileInodeData.volStruct.volumeName, inFlightFileInodeData.InodeNumber)
-
 	select {
 	case flushFirst = <-inFlightFileInodeData.control:
 		// All we needed was the value of flushFirst from control chan
@@ -310,6 +312,8 @@ func (mS *mountStruct) doInlineCheckpoint() {
 	var (
 		err error
 	)
+	logger.Tracef("fs.doInlineCheckpoint() entry: volume '%s'", mS.volStruct.volumeName)
+	defer logger.Tracef("fs.doInlineCheckpoint() return: volume '%s'", mS.volStruct.volumeName)
 
 	if nil == mS.headhunterVolumeHandle {
 		mS.headhunterVolumeHandle, err = headhunter.FetchVolumeHandle(mS.volStruct.volumeName)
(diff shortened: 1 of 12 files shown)


proxyfsd/default.conf
```diff
@@ -97,9 +97,19 @@ Debug:           false
...
+
+# Enable trace logging on a per-package basis. Trace logs are disabled by default unless enabled here.
+# Supported values: any combination of "dlm", "fs", "fuse", "headhunter", "inode", "jrpcfs", "logger",
+# and "swiftclient" or "none" (default).
+TraceLevelLogging: none
+
+# Enable debug logging on a per-package basis. Debug logs are disabled by default unless enabled here.
+# Supported values: any combination of "ldlm", "fs", "jrpcfs" and "inode" or "none" (default).
+DebugLevelLogging: none
+
```
edmc-ss, line 112: Seems to me this list is very dated... shouldn't we update it?

proxyfsd/logging.conf
```diff
@@ -4,7 +4,8 @@ LogFilePath: proxyfsd.log
 # NOTE: Log levels other than Trace and Debug are always on.
 
 # Enable trace logging on a per-package basis. Trace logs are disabled by default unless enabled here.
-# Supported values: jrpcfs, inode, none (default).
+# Supported values: any combination of "dlm", "fs", "fuse", "headhunter", "inode", "jrpcfs", "logger",
+# and "swiftclient" or "none" (default).
```
edmc-ss, line 8: Same comment as above... the list of packages may have grown since this was put together...

proxyfsd/saio_logging.conf
```diff
@@ -4,7 +4,8 @@ LogFilePath: /var/log/proxyfsd/proxyfsd.log
 # NOTE: Log levels other than Trace and Debug are always on.
 
 # Enable trace logging on a per-package basis. Trace logs are disabled by default unless enabled here.
-# Supported values: jrpcfs, inode, none (default).
+# Supported values: any combination of "dlm", "fs", "fuse", "headhunter", "inode", "jrpcfs", "logger",
+# and "swiftclient" or "none" (default).
```
edmc-ss, line 8: Again...

edmc-ss:
In general, I am no fan of "let's just log/trace everything"... but can be comfortable with it *IF-AND-ONLY-IF* the tracing is low overhead (particularly when disabled). The current implementation is not alas. So I'm quite nervous about merging this in...

inode/file.go
```diff
@@ -617,6 +657,7 @@ func (vS *volumeStruct) Write(fileInodeNumber InodeNumber, offset uint64, buf []
 	fileInode.AttrChangeTime = updateTime
 	fileInode.ModificationTime = updateTime
 	fileInode.NumWrites++
+	fileInode.dirty = true
```
bschatz-swift, line 660: I am curious - did you have a specific test which was failing that caused you to find this bug?
Or, simply by reading the code did you notice this was not set correctly?

fs/api_internal.go
```diff
@@ -170,13 +170,15 @@ func (vS *volumeStruct) inFlightFileInodeDataFlusher(inodeNumber inode.InodeNumb
 }
 
 func (inFlightFileInodeData *inFlightFileInodeDataStruct) inFlightFileInodeDataTracker() {
-	var (
-		flushFirst bool
-	)
 
-	logger.Tracef("fs.inFlightFileInodeDataTracker(): waiting to flush volume '%s' inode %d",
+	logger.Tracef("fs.inFlightFileInodeDataTracker() entry: waiting to flush volume '%s' inode %d",
```
bschatz-swift, line 174: Logger already adds the function name to the log message so you should not have to add the function name here.

bschatz-swift: For example,  here is an example log message that I see when I am using this PR:

time="2017-12-11T21:00:11.737919Z" level=info msg="vs.Write(): growing file size 999424 to 1000000  volume 'CommonVolume'  inode 204  offset 999424  len 576" function=Write goroutine=981 package=inode

inode/file.go
bschatz-swift: Oh - I see you just moved it down further with the rest of the changes.

PR #61, opened 2017-12-14
Check constraints when PUT-ing and POST-ing objects

tipabu:


smerritt:
This helps, but there's more to it. For example, directory names in object paths can have at most 255 characters in them, but I don't believe the middleware is currently enforcing this.

tipabu:
Something worth remembering -- we have two fundamentally different types of constraints that we're in charge of checking:

1. all the "normal" Swift constraints, because we're acting like a proxy-server, and
2. all the *new* constraints that are imposed by ProxyFS, because it's a filesystem.

This addresses just the first kind; the directory-name-length limit is an example of the second.

I'll rebase and add a couple more patches.

PR #62, opened 2017-12-20
POST should update Last-Modified

tipabu:
Might be better to just *always* use `StatCTime` -- is `StatCTime >= StatMTime` necessarily true?

### DIFF
--- a/fs/api.go
+++ b/fs/api.go
@@ -27,6 +27,7 @@ type ReadRangeIn struct {
 type AccountEntry struct {
 	Basename         string
 	ModificationTime uint64 // nanoseconds since epoch
+	AttrChangeTime   uint64 // nanoseconds since epoch
 }
 
 // Returned by MiddlewareGetContainer
@@ -35,6 +36,7 @@ type ContainerEntry struct {
 	Basename         string
 	FileSize         uint64
 	ModificationTime uint64 // nanoseconds since epoch
+	AttrChangeTime   uint64 // nanoseconds since epoch
 	IsDir            bool
 	NumWrites        uint64
 	InodeNumber      uint64
@@ -45,6 +47,7 @@ type HeadResponse struct {
 	Metadata         []byte
 	FileSize         uint64
 	ModificationTime uint64 // nanoseconds since epoch
+	AttrChangeTime   uint64 // nanoseconds since epoch
 	IsDir            bool
 	InodeNumber      inode.InodeNumber
 	NumWrites        uint64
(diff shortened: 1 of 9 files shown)


smerritt:
yeah, ctime >= mtime always IIRC

tipabu:
Looks like `AttrChangeTime` is always set by proxyfs, whereas `ModificationTime` can be user-settable. There are even some [comments over in api_internal.go](https://github.com/swiftstack/ProxyFS/blob/1.0.3/fs/api_internal.go#L2640-L2648) about it. The more I think about this, the more I like changing _all_ references in the middleware APIs to use `AttrChangeTime`.

And to help track the plumbing (since I know it took *me* a bit):

- `(mS *mountStruct) MiddlewarePost` [calls `PutStream`](https://github.com/swiftstack/ProxyFS/blob/1.0.3/fs/api_internal.go#L1696)
- `(vS *volumeStruct) PutStream` [updates `AttrChangeTime`](https://github.com/swiftstack/ProxyFS/blob/1.0.3/inode/inode.go#L1280-L1281) but doesn't touch `ModificationTime` (or even `AccessTime`, which seems... suspicious).

smerritt:
I'm not sure what to do with this change.

On the one hand, I don't like having `ModificationTime` in the RPC responses be ctime and not mtime. It'd be better for future comprehensibility if we made the RPC responses contain `ChangeTime` instead.

Oh the other hand, I don't like what that proposed RPC change does to running clusters during a rolling upgrade. Old middleware / new proxyfsd will break looking for `ModificationTime`, while new middleware / old proxyfsd will break looking for `ChangeTime`.

On the gripping hand, we don't have a good plan for dealing with rolling cluster upgrades with ProxyFS, and I think we need to figure something out there, and then we'll know what to do here.

tipabu:
How about this as a plan:
- Leave `ModificationTime` alone -- it continues to mean what it has always meant.
- Add a new `ChangeTime` field to the existing responses/response-entries that need it.
- Update the middleware to use `ChangeTime` if present, but fall back to `ModificationTime` if not.

smerritt:
That sounds like a good plan. We should also keep track of protocol cruft somewhere so that we can get rid of it someday, perhaps after a checkpoint release.

tipabu:
OK, are there any timestamps stored that let us see when a directory was created or had its metadata updated, but that *wouldn't* be affected by new files being created under it? I'm looking at `test.functional.tests.TestContainer.testContainerLastModified`; in particular, [the latter half](https://github.com/openstack/swift/blob/2.16.0/test/functional/tests.py#L896-L909) is currently failing (at least, with this change; maybe before too).

The good news is, I *have* managed to get runway working for me now, and can confirm that this PR fixes `test.functional.tests.TestFile.test_POST`.

smerritt:
Addressing the "XXX WTF" comments: looks like that code got pulled out into a helper function at some point, and the function name didn't get changed. It probably should be.

@edmc-ss perhaps you could address the question of "are there any timestamps stored that let us see when a directory was created or had its metadata updated, but that wouldn't be affected by new files being created under it?" I can't think of any off the top of my head.

Other than that, it looks good to me.

tipabu:
> Addressing the "XXX WTF" comments:

Heh, whoops. Forgot about that running commentary... Will remove before merge.

tipabu:
@smerritt removed the comments.

@edmc-ss (or anyone else) any other thoughts on this? Or should we just go ahead and merge it?

edmc-ss:
I added Craig & myself to the review of this. I'd like to get this in asap but want to get Craig to (also) weigh in her since he's got a parallel set of permissions fixes going in for the NFS side of things as well...

tipabu:
Rebased to resolve merge conflicts.

charmster:
@tipabu files and directories in ProxyFS have a CreationTime that is set when the file or directory is created and never updated subsequently.  Note that CreationTime is not part of POSIX but some LInux file systems support it (called "Birth" in the stat(1) command, but i'm not sure which system call returns it).

`st_ctime` (AttrChangeTime) is set to the current time any time file data or metadata is updated, except for the update to `st_atime` when the file is read.  `st_mtime` is set to the current time when the file is created, written to, or has its size changed (including writing via a file mapping).

howver, both `st_atime` and `st_mtime` can be set via the `utimes(2)` system call, which means they could have any value.   note that `st_ctime` is set to the current time as a result, so its really the only thing you can trust to know if the file has changed.

`st_ctime` is monotonically increasing so long as the system clock is monotonically increasing.  but if somebody decides to set the system clock backwards by 1 year, then `st_ctime` can go "backward" when a file is updated.

there's no timestamp on the directory that does what you want.

charmster (approved):
Sorry i was so slow to look at this.  Yes, this change seems fine.  I trust AttrChangeTime (ctime) much more then mtime (Modification time).

PR #65, opened 2018-01-04
Start and stop Swift processes cleanly and FUSE hang

bschatz-swift:
During CHO testing, we noticed that Swift processes were not always
started and stopped as expected when using start_and_mount_pfs or
unmount_and_stop_pfs.  Even though we stopped Swift there were Swift
processes which were still running.  Additionally, the Swift reaper
process was not enabled and therefore the checkpoint container
was not being cleaned up.

The first part of the fix is to do "swift-init all start" and
"swift-init all stop" for starting and stopping Swift.  This also
starts the Swift reaper.

The second part of the fix is in the functional-tests repository.  It is
to have the kill_proxyfs.sh script do a pkill and kill remaining Swift
processes which exist.

### DIFF
--- a/cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
+++ b/cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
@@ -55,7 +55,10 @@ if [ -f /usr/bin/systemctl ]; then
     # sudo /usr/bin/systemctl stop winbind
     sudo /usr/bin/systemctl stop proxyfsd
     await_proxyfsd_shutdown
-    sudo /usr/bin/swift-init main stop
+
+    # NOTE: We purposely stop Swift with "swift-init all stop" and not
+    # "swift-init main stop".   See start_and_stop_pfs for more details.
+    sudo /usr/bin/swift-init all stop
     # sudo /usr/bin/systemctl stop swift
     sudo /usr/bin/systemctl stop memcached
 else
@@ -69,7 +72,10 @@ else
     # sudo /usr/sbin/service winbindd stop
     sudo /usr/sbin/service proxyfsd stop
     await_proxyfsd_shutdown
-    sudo /usr/bin/swift-init main stop
+
+    # NOTE: We purposely stop Swift with "swift-init all stop" and not
+    # "swift-init main stop".   See start_and_stop_pfs for more details.
+    sudo /usr/bin/swift-init all stop
     # sudo /usr/sbin/service swift stop
     sudo /usr/sbin/service memcached stop
 fi
--- a/cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
+++ b/cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
@@ -59,7 +59,6 @@ if [ -f /usr/bin/systemctl ]; then
     # NOTE: We purposely stop Swift with "swift-init all stop" and not
     # "swift-init main stop".   See start_and_stop_pfs for more details.
     sudo /usr/bin/swift-init all stop
-    # sudo /usr/bin/systemctl stop swift
     sudo /usr/bin/systemctl stop memcached
 else
     # Ubuntu (not tested!)
@@ -76,6 +75,5 @@ else
     # NOTE: We purposely stop Swift with "swift-init all stop" and not
     # "swift-init main stop".   See start_and_stop_pfs for more details.
     sudo /usr/bin/swift-init all stop
-    # sudo /usr/sbin/service swift stop
     sudo /usr/sbin/service memcached stop
 fi
--- a/cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
+++ b/cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
@@ -56,9 +56,7 @@ if [ -f /usr/bin/systemctl ]; then
     sudo /usr/bin/systemctl stop proxyfsd
     await_proxyfsd_shutdown
 
-    # NOTE: We purposely stop Swift with "swift-init all stop" and not
-    # "swift-init main stop".   See start_and_stop_pfs for more details.
-    sudo /usr/bin/swift-init all stop
+    sudo /usr/bin/swift-init main stop
     sudo /usr/bin/systemctl stop memcached
 else
     # Ubuntu (not tested!)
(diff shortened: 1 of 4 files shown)


cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
```diff
@@ -55,7 +55,10 @@ if [ -f /usr/bin/systemctl ]; then
     # sudo /usr/bin/systemctl stop winbind
     sudo /usr/bin/systemctl stop proxyfsd
     await_proxyfsd_shutdown
-    sudo /usr/bin/swift-init main stop
+
+    # NOTE: We purposely stop Swift with "swift-init all stop" and not
+    # "swift-init main stop".   See start_and_stop_pfs for more details.
+    sudo /usr/bin/swift-init all stop
     # sudo /usr/bin/systemctl stop swift
```
bschatz-swift: Can I remove the above line "# sudo /usr/bin/systemctl stop swift" ?

cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
```diff
@@ -69,7 +72,10 @@ else
     # sudo /usr/sbin/service winbindd stop
     sudo /usr/sbin/service proxyfsd stop
     await_proxyfsd_shutdown
-    sudo /usr/bin/swift-init main stop
+
+    # NOTE: We purposely stop Swift with "swift-init all stop" and not
+    # "swift-init main stop".   See start_and_stop_pfs for more details.
+    sudo /usr/bin/swift-init all stop
     # sudo /usr/sbin/service swift stop
```
bschatz-swift: Can I remove the above line "# sudo /usr/sbin/service swift stop" ?

cookbooks/proxyfs/templates/default/usr/bin/start_and_mount_pfs.erb
```diff
@@ -134,13 +143,13 @@ if [ -f /usr/bin/systemctl ]; then
     # sudo /usr/bin/systemctl stop winbind
     sudo /usr/bin/systemctl stop proxyfsd
     await_proxyfsd_shutdown
-    sudo /usr/bin/swift-init main stop
+    sudo /usr/bin/swift-init all stop
     # sudo /usr/bin/systemctl stop swift
```
bschatz-swift: Can I remove all lines in this section which look like "# sudo /usr/bin/systemctl stop swift" i.e. all lines that have swift commented out?

bschatz-swift (commented):
Can I make similar changes to saio/bin/start_and_mount_pfs and saio/bin/unmount_and_stop_pfs ?

bschatz-swift:
@edmc-ss  Can I make similar changes to saio/bin/unmount_and_stop_pfs and saio/bin/start_and_mount_pfs ?

cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
gerardgine: I think we expected to be able to use that line at some point, when we had properly set it up. Now I'm starting to think it won't happen in the near future, so yeah, get rid of it. Also, it's not that hard to re-write it. ;)

cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
gerardgine: Yes

cookbooks/proxyfs/templates/default/usr/bin/start_and_mount_pfs.erb
gerardgine: Yes

gerardgine (approved):
I have never used swift-init all start/stop, but if your tests show it works better, I'm all in.

cookbooks/proxyfs/files/default/usr/bin/unmount_and_stop_pfs
```diff
@@ -69,7 +71,9 @@ else
     # sudo /usr/sbin/service winbindd stop
     sudo /usr/sbin/service proxyfsd stop
     await_proxyfsd_shutdown
-    sudo /usr/bin/swift-init main stop
-    # sudo /usr/sbin/service swift stop
+
+    # NOTE: We purposely stop Swift with "swift-init all stop" and not
+    # "swift-init main stop".   See start_and_stop_pfs for more details.
```
gerardgine: Not very important, but I guess you mean `start_and_mount_pfs` instead of `start_and_stop_pfs`.

gerardgine (approved):
Looks good to me, but I only checked the orchestration scripts, not the go code.

charmster:
It looks pretty good.  I'm glad that proxyfsd is no longer doing a stat(2) on the FUSE mount point!

i'm not sure why we moved from using `systemctl` to manage swift to running the swift directly:
```
-    # sudo /usr/bin/systemctl start swift
￼ -    sudo /usr/bin/swift-init main start
￼ +    sudo /usr/bin/swift-init all start
```
i was going to comment that it would be better to edit the systemctl config file at: `/usr/lib/systemd/system/swift.service` to change:
```
[Service]
Environment=NSS_SDB_USE_CACHE=YES
Environment=PATH=/usr/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/sbin:/bin
- ExecStart=/usr/bin/swift-init main start
+ ExecStart=/usr/bin/swift-init all start
- ExecReload=/usr/bin/swift-init main stop
+ ExecReload=/usr/bin/swift-init all stop
```
but you weren't the person that moved us away from systemctl.

i may have found an explanation for our extra processes: on my (old) runway environment -- systemctl seems to be starting Swift automatically.  i found i needed to use `sudo systemctl stop swift; sudo systemctl disable swift` to make it stop.  otherwise, after `unmount_and_stop_pfs` there were still a bunch of processes running.

you can use `sudo sysemctl status swift` to find out what systemctl thinks is going on.

(if you do play with the systemctl config files you need to run `sudo systemctl daemon_reload` after modifying them so it will pick them up.)

i also find that the difference between `swift-init main start` and `swift-init all start` is very significant in terms of CPU consumption.  i'm maxing out the 2 CPU that are assigned to the VM and getting throttled by the CPU.  i need to figure out how to give the VM 2 more CPU.

PR #67, opened 2018-01-25
Jrpcfs shutdown fix

balajisrao:


jrpcfs/config.go
```diff
@@ -33,6 +35,12 @@ type globalsStruct struct {
 
 	// Map used to store volumes already mounted for bimodal support
 	bimodalMountMap map[string]fs.MountHandle
+
+	// Connection list and listner list to close during shutdown:
```
edmc-ss: ^listner^listener

jrpcfs/config.go
```diff
@@ -33,6 +35,12 @@ type globalsStruct struct {
...
+	connLock      sync.Mutex
+	connections   *list.List
+	connWaitGroup sync.WaitGroup
+	listners      []net.Listener
```
edmc-ss: !!

jrpcfs/config.go
```diff
@@ -115,6 +123,8 @@ func Up(confMap conf.ConfMap) (err error) {
 		}
 	}
 
+	globals.connections = list.New()
```
edmc-ss: So you are just using one connections list to cover both types of connections?

edmc-ss: And don't you also need to globals.listners = make([]net.Listener)?

jrpcfs/filesystem.go
```diff
@@ -20,7 +20,6 @@ import (
 
 // RPC server handle
 var srv *rpc.Server
```
edmc-ss, line 22: I think these three should now be local and in .globals... not here, right?

jrpcfs/io.go
```diff
@@ -41,18 +46,27 @@ func ioServerLoop() {
...
 
-		go ioHandle(conn)
+		globals.connWaitGroup.Add(1)
+
+		globals.connLock.Lock()
+		globals.connections.PushBack(conn)
+		globals.connLock.Unlock()
+
+		go func() {
+			ioHandle(conn)
```
edmc-ss: Don't we need a:
```
globals.connLock.Lock()
globals.connections.Remove(conn)
globals.connLock.Unlock()
```
here?

jrpcfs/filesystem.go
```diff
@@ -53,17 +57,26 @@ func jrpcServerLoop() {
...
 
-		go srv.ServeCodec(jsonrpc.NewServerCodec(conn))
+		globals.connWaitGroup.Add(1)
+
+		globals.connLock.Lock()
+		globals.connections.PushBack(conn)
+		globals.connLock.Unlock()
+
+		go func() {
+			srv.ServeCodec(jsonrpc.NewServerCodec(conn))
```
edmc-ss: Don't we need a:
```
globals.connLock.Lock()
globals.connections.Remove(conn)
globals.connLock.Unlock()
```
here?

edmc-ss (changes requested):
Looks pretty close...
My only technical suggestion was to ensure you remove conn's from the connections list as their goroutines exit...

jrpcfs/config.go
```diff
@@ -299,5 +309,16 @@ func Down() (err error) {
 	err = nil
 	jsonRpcServerDown()
 	ioServerDown()
+
+	globals.Lock()
+	for _, listner := range globals.listners {
+		listner.Close()
+	}
+
+	for elm := globals.connections.Front(); elm != nil; elm = elm.Next() {
```
edmc-ss: Ah... and per my earlier comment about this connections list being updated as each goroutine exits... this looping would need to be careful to lock/unlock around accesses...

The way you've done it... you are just sending close's... to a lock/unlock around the loop is fine.

jrpcfs/config.go
```diff
@@ -299,5 +309,16 @@ func Down() (err error) {
...
+		conn := elm.Value.(net.Conn)
+		conn.Close()
+	}
+	globals.Unlock()
```
edmc-ss: And, finally, I expected to see a globals. connWaitGroup.Wait() call here, right?

edmc-ss (changes requested):
Ok - I had another technical set of comments... basically I was looking for the Wait() call to await both the Listeners and all the child goroutines' completions.

jrpcfs/config.go
balajisrao: Done!

jrpcfs/config.go
balajisrao: Done

jrpcfs/io.go
balajisrao: YES. fixed it.

jrpcfs/filesystem.go
balajisrao: YES

jrpcfs/config.go
balajisrao: YES - fixed it.

jrpcfs/config.go
balajisrao: yes, that way the list is consistent till all close() calls are done and then the list becomes empty as each connections close.

edmc-ss (approved):
With the make([]net.Listener, 2) changed to make([]net.Listener, 0, 2), this looks perfect!

PR #68, opened 2018-01-26
Removing test targets so deb builder doesn't break

FoxLisk:


edmc-ss (approved):
This looks perfect... I agree... no need to run those tests + particularly the c-test rule might not be valid anyway.

PR #69, opened 2018-01-30
Pinned most packages Runway & SAIO install @ provisioning

gerardgine:
For now, I've pinned all of the packages Runway installs on RHEL and most of the packages SAIO installs.


edmc-ss:
Not exactly sure we need to pin tools like atop & wireshark... but it's ok.

I'll merge this after I verify provisioning SAIO for ProxyFS works as well...

PR #72, opened 2018-02-09
Work better with cloud sync

tipabu:


tipabu:
I'm gonna break this up a bit; expect a couple new PRs soon.

tipabu:
I haven't actually *seen* the bad that I was expecting when middlewares were in the wrong order -- maybe this isn't necessary? But it *seems like* it should be...

tipabu:
I'm going to assume this isn't actually necessary.

PR #73, opened 2018-02-09
Work with new Swift

tipabu:
Here we

* Add `X-Timestamp` headers before calling `constraints.check_object_creation` (like the object controller had to do in https://github.com/openstack/swift/commit/31c294d),
* Update the version of Swift we use when running unit tests to demonstrate why the above is necessary,
* Add more tox envs to test with both older and newer versions of Swift as well, which highlights why we can't use things like `Timestamp.now()`, and
* Add the pfs_middleware unit tests to the project-wide `make test` target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,8 @@ test:
 	@set -e; \
 	for gosubdir in $(gosubdirs); do \
 		$(MAKE) --no-print-directory -C $$gosubdir test; \
-	done
+	done; \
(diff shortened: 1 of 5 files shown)


smerritt (approved):
Looks good to me.

I sort of want CI to test against every Swift version we care about, but that can be a different set of changes.

edmc-ss:
I just pushed a small delta to Tim's PR that makes sure runway & SAIO can complete "make"...

Everything else looks great!

PR #74, opened 2018-02-13
pfs_middleware: omit metadata, manifest headers from log segments

smerritt:
When uploading an object, its data is split into one or more log
segments by pfs_middleware. Previously, we were skipping certain
dangerous headers (X-Delete-At, X-Delete-After, and ETag) that could
cause the PUT of a log segments to fail, but we were letting pretty
much every other header pass through unscathed. This made things a bit
awkward for any client that examined our log segments directly, as
some log segments may have X-Static-Large-Object (if a client uploaded
an SLO manifest), X-Object-Manifest (DLO), arbitrary sysmeta
(symlinks, ACLs, third-party middlewares), and arbitrary user
metadata.

Now, instead of blacklisting certain headers, we whitelist the ones we
want. For each log segment, we now construct a new, blank swob.Request
object and populate the minimal set of attributes required to succeed
in making the PUT request.

Bonus fix: cleaned up some code for detecting the end of client input
on a PUT request. We had two different methods for checking, and now
we're down to just one.

### DIFF
--- a/pfs_middleware/pfs_middleware/middleware.py
+++ b/pfs_middleware/pfs_middleware/middleware.py
@@ -122,8 +122,6 @@ ZERO_FILL_PATH = "/0"
 
 LEASE_RENEWAL_INTERVAL = 5  # seconds
 
-FORBIDDEN_OBJECT_HEADERS = {"X-Delete-At", "X-Delete-After", "Etag"}
-
 ORIGINAL_MD5_HEADER = "X-Object-Sysmeta-ProxyFS-Initial-MD5"
 
 # They don't start with X-Object-(Meta|Sysmeta)-, but we save them anyway.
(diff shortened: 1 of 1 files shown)


pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -553,8 +546,6 @@ def read(self, length=None, *args, **kwargs):
 
         self.bytes_read += len(chunk)
         self.limit -= len(chunk)
-        if len(chunk) == 0:
-            self.hit_eof = True
         return chunk
 
     def readline(self, size=None, *args, **kwargs):
```
tipabu, line 551: Off topic: This doesn't always respect the readline contract. In particular, if `self._peeked_data == '\n'` then blank lines can cause trouble. For example
```
>>> from pfs_middleware.middleware import LimitedInput
>>> fp = open('tox.ini')
>>> li = LimitedInput(fp, 1000)
>>> li.readline(), li.has_more_to_read
('[tox]\n', True)
>>> li.readline(), li.has_more_to_read
('envlist = py35,py27,pep8\n', True)
>>> li.readline(), li.has_more_to_read
('minversion = 1.6\n', True)
>>> li.readline(), li.has_more_to_read
('skipsdist = True\n', True)
>>> li.readline(), li.has_more_to_read
('\n[testenv]\n', True)  # has two lines!
```

tipabu (approved):
Makes sense, and the swift func tests seem about as unhappy as they usually are. Manually verified that we no longer send object headers down to the log segment data.

PR #77, opened 2018-02-17
Initial version of parallelized FSCK

edmc-ss:
The TODO section at the end of validateVolume() in fs/treewalk.go
defines the current view of remaining work.

As for as how this new version matches up to the functionality it
is replacing, only the LinkCount adjustment has not been applied.


edmc-ss:
This version of FSCK is by no means complete... lots of TODOs remain as to what I'm hoping to do.
I just wanted to get this out there for others to see what's going on in the code base re FSCK.
The key change is there is this "goroutine" strategy to launch at every opportunity.
Each inode is validated in a distinct goroutine. Each dir found during treewalk gets its own goroutine.
To limit parallelism, there is a "gate" / "turnstile" per parallelized function arbitrarily tuned at present.

Please let me know what you think - comments are welcome!

edmc-ss:
Ok - getting there...

The remaining TODOs (10) are all beyond what the original FSCK performed.
Hence, this PR technically could be merged at any point moving forward.
The benefit is parallelism... and providing the framework for the remaining work.

But I'll just keep working on those TODOs one by one in the meantime.

Review comments very welcome!

edmc-ss:
This is just a call out to my three invited reviewers to, uh, review my PR :-)

PR #78, opened 2018-02-21
Update bimodal support docs

tipabu:
I *think* this gives better/more targeted answers about what to expect given the various read and write paths.


docs/source/api/index.rst
```diff
@@ -2,344 +2,138 @@
...
+  ``?multipart-manifest=get`` query parameter via the object API.
+- Swift "symlink" objects are only followed when accessed through the object
+  API. When accessed through the filesystem API, these will be zero-byte
+  files.
+
+Reading files via the object API
+--------------------------------
+
+- Since containers map to directories, files written at the root of the
+  volume are not accessible via the object API.
```
tipabu, line 100: This maybe deserves a note about snapshots -- if you take a snapshot, you can get to the root-level file via the snapshot.

docs/source/api/index.rst
```diff
@@ -2,344 +2,138 @@
...
+  <https://tools.ietf.org/html/rfc2616.html#section-13.3.3>`__ and
+  `RFC 7232 <https://tools.ietf.org/html/rfc7232#section-2.3>`__
+  intended.
+
+.. note::
+  Filesystem access to SLO segment data can invalidate the
+  manifest, as the segment ETag may change.
+
+
+Swift functionality broken as of 1.3
```
tipabu, line 114: Should probably update the version number. As of 1.9, all of these are still broken (to my knowledge).

PR #85, opened 2018-02-26
Bugfix object leaks

charmster:
File systems that are around for awhile seem to leak objects -- the amount of space used in terms of objects seems to keep growing.  If all files and directories are deleted, there can still be a lot of objects left over.

It turns out there were two bugs in the sortedmap package that were fixed in the following commits:

d64c276 -- when rebalanceHere() discards "rightSiblingNode" as part of merging two nodes, it fails to mark it dirty which means the underlying disk space is not reqcovered

a594a01 -- discardNode() stops walking the tree when it finds a node that is not loaded which means that tree.BPlusTreeCallbacks.DiscardNode() is not called to recover the underlying disk space

the rest of the commits simply add, remove, or update tracing of Swift objects and tree nodes.  They can be ignored (unless you're interested in that type of thing).


proxyfsd/saioproxyfsd0.conf
```diff
@@ -13,6 +13,8 @@ WhoAmI: Peer0
 .include ./httpserver.conf
 .include ./debug.conf
 
+.include ./proxyfs-evtlog.conf
+
 # put it here to have the last word (override previous)
-#[Logging]
-#TraceLevelLogging: inode fs headhunter
+[Logging]
+TraceLevelLogging: headhunter inode sortedmap
```
edmc-ss: I assume this was because you had instrumented sortedmap...
All those traces are gone now (good thing), so I guess we could delete "sortedmap" here, right?

vendor/github.com/swiftstack/sortedmap/btree.go
```diff
@@ -83,12 +83,12 @@ type onDiskNodeStruct struct {
 	//                                  counted <N-1> number of Key:onDiskReferenceToNodeStruct pairs
 }
 
-type onDiskReferenceKeyStruct struct { // Used as Key for staleOndiskReferences LLRBTree below
+type onDiskReferenceKeyStruct struct { // Used as Key for staleOnDiskReferences LLRBTree below
```
edmc-ss, line 86: I'm just gonna put this comment here to remind us that these changes actually need to find there way into the corresponding file in the actual sortedmap repo. Here, you've changes the vendorized "copy" of the file that would, of course, be overwritten the next "glide up".

I'll take care of copying them over along with other Dispose() misses we talked about this morning I found inspired by you. Indeed, I now recall that I never really finished the Dispose() callback code - it came along back when we were looking at GC activities at many levels... and I believe we identified sortedmap as needing serious completion work that never happened. My oops!

vendor/github.com/swiftstack/sortedmap/btree.go
```diff
@@ -1232,6 +1236,31 @@ func (tree *btreeTreeStruct) insertHere(insertNode *btreeNodeStruct, key Key, va
 	return
 }
 
+func (tree *btreeTreeStruct) AssertNodeEmpty(node *btreeNodeStruct) {
```
edmc-ss, line 1239: Cool function - but we might want to return err to the caller in this case.
That is the pattern adopted elsewhere i.e. non-nil err always means corruption.
Where "errors" are "occasionally ok", I've also returned an ok bool e.g. when the caller does a Get on a non-existent key.

vendor/github.com/swiftstack/sortedmap/btree.go
```diff
@@ -1656,6 +1705,14 @@ func (tree *btreeTreeStruct) flushNode(node *btreeNodeStruct, andPurge bool) (er
 
 	if node.dirty {
 		tree.postNode(node) // will also mark node clean/used in LRU
+
+		// if this node is dirty the parent must be as well
+		if node.parentNode != nil && !node.parentNode.dirty {
+			err = fmt.Errorf("flushNode(): tree %p node %p: parent node %p not marked dirty",
+				tree, node, node.parentNode)
+			panic(err)
```
edmc-ss, line 1713: See earlier comment about sortedmap returning non-nil err rather than panic'ing...

edmc-ss (changes requested):
The more general statement about panic'ing out of sortedmap kind of goes against my gut in that corruption due to logic failures should stop as soon as possible - but I deferred to others on the team (at the time) who preferred pushing the corruption indication upwards.

vendor/github.com/swiftstack/sortedmap/btree.go
charmster: Apparently I haven't fixed all of the problems yet!  I just deleted my "huge" (2 Gibyte) file system and there are lots of __checkpoint__ objects still there.  Investigating ...

vendor/github.com/swiftstack/sortedmap/btree.go
charmster: I actually think that's a bad pattern!  If the tree is now corrupt _due to an internal logic error in the code_ then returning an error to the caller seems pointless.  What is the caller supposed to do???  The tree should not be accessed again.  Certainly it will be difficult to debug the situation with all the context lost (code at the Proxyfs layer looking at an error return).

Note that it is possible to catch a panic so a caller could catch this type of error if they wanted to (and log a useful error message).  I suppose a caller that wanted to recover could drop the tree on the ground and try starting over with a new tree ... though i don't think Proxyfs would be able to do that.

I do draw a distinction between internal logic errors and errors caused for external reasons, such as a bad value passed by the caller or being unable to read a node from disk (which _will happen_ and is not the fault of the the sortedmap code).  But, as near as i can think of failing to read a node from disk will not result in a corrupt tree, it just means that some nodes are inaccessible (either temporarily or permanently).  I think continuing to use the rest of the tree should be fine.  Having part of the tree inaccessible is a different situation then, say, inserting a new key and getting half way up the tree splitting nodes only to discover that some parent node can no longer perform Len().

This would require some careful thought about nodes that are readable from disk but turn out to be corrupt.  I don't think we want to panic in that situation.

I gave some thought to converting all those error cases that should never happen into panic()s, but decided to go with the smaller change and discuss this issue later (like now!).

charmster:
Apparently I haven't fixed all of the problems yet! I just deleted my "huge" (2 Gibyte) file system and there are lots of checkpoint objects still there. Investigating ...

Nevermind ... it was just that the container server hadn't caught up with all the object deletes so it was still showing some objects present that did not actually exist.

vendor/github.com/swiftstack/sortedmap/btree.go
charmster: I need to look at your other Dispose() fixes.  apparently i haven't fixed all the problems after all.

vendor/github.com/swiftstack/sortedmap/btree.go
edmc-ss: You and I are on the same page w.r.t. logic errors... but others have a different opinion.
I'd rather make a decision on the pattern and implement it consistently. I'm merely saying that your change makes this pattern inconsistent.

Let's chat about it tomorrow - I think it's a good discussion to have.

edmc-ss:
I see that your latest commit did, at least, pass the "overlord that is travis" CI tests :-)

PR #93, opened 2018-03-02
Updated Runway & SAIO to use pre-built Samba headers

gerardgine:


gerardgine:
@edmc-ss I've updated the provisioning file for the SAIO machine. That script already does lots of checks that we don't do in either Runway or TravisCI, so basically, I only changed the name of the directory where the pre-built headers live. Can you have a look before I merge into development?

edmc-ss (approved):
Looks great!

PR #95, opened 2018-03-05
Added "SkipRetry" option for Object DELETEs in package swiftclient

edmc-ss:
In particular, now that we have a garbage collection step in FSCK
(admittedly, not yet fulfilling this promise for File Inode LogSegment
objects), we can move to no-retry on object DELETEs thus significantly
speeding up that very FSCK operation that may find a need to DELETE
objects that are already deleted. Should the lack of retry end up
missing a DELETE on an existing object, the subsequent FSCK would
catch it and retry then.


edmc-ss:
Massive improvement in FSCK performance :-)

PR #98, opened 2018-03-13
Fixed remaining scalability problems with FSCK

edmc-ss:
Now, the "grab parallelism" step is performed *BEFORE* launching a
goroutine. Previously, the goroutine was used to grab it. This would
be an ok pattern if one is merely trying to limit the number of
such goroutines executing in parallel because too many would be
inefficient... but here, we are actually worried about the sheer
number of goroutines being too great such that just holding that
number of goroutine contexts in (virtual) memory a potential challenge.

Also:
  Added "End Time" in FSCK/SCRUB page table.
  Updated Layout Report (both JSON & HTML)
  Integrated Trigger pages into JS-enabled single page

Finally, pfs-crash now performs the SCRUB *BEFORE* the FSCK since
the SCRUB may have deleted Inodes that are now not found during the
subsequent FSCK.

### DIFF
--- a/fs/treewalk.go
+++ b/fs/treewalk.go
@@ -33,6 +33,8 @@ const (
 
 	scrubVolumeInodeParallelism = uint64(100)
 
+	invalidLinkCount = uint64(0xFFFFFFFFFFFFFFFF) // Indicates Inode to be removed
+
 	lostAndFoundDirName = ".Lost+Found"
 )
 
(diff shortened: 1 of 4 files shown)


gerardgine:
I don't know about the changes in the FSCK tool itself, but I love the new web pages. ;)

gerardgine (approved):

PR #118, opened 2018-04-30
Snapshot phase 2

edmc-ss:
The SnapShot Phase 2 branch has now achieved parity with what's on Development,
so it should be safe to merge in. At this point, checkpoints are still in V2 format. But
much of the plumbing is in place to move on to the next phase in which SnapShots will
actually be maintained (Phase 3). A final phase (Phase 4) will actually make them
visible via the /.snapshot/ directory path By Name (i.e. inside the /.snapshot/ directory,
each of the active SnapShots will be listed By Name as subdirectories of /.snapshot/...
with the contents of each of those subdirectories matching the corresponding contents
of / at the time each SnapShot was created).

### DIFF
--- a/fs/api.go
+++ b/fs/api.go
@@ -5,8 +5,6 @@ package fs
 import "C"
 
 import (
-	"time"
-
 	"github.com/swiftstack/ProxyFS/inode"
 	"github.com/swiftstack/ProxyFS/stats"
 	"github.com/swiftstack/ProxyFS/utils"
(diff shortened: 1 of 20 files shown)


edmc-ss:
Hmm... Intermittent go test inode failures persist... gonna fix that first before merging :-(

edmc-ss:
Ok... so I cannot reproduce this when doing clean builds... so that was probably the problem.|

Merging...

PR #125, opened 2018-05-23
Jrpc fixes

charmster:
i noticed a typo in the recent checkin that should be fixed or else we may not always pass the right connection to a Go routine.


charmster:
Need to do redo this with less fixes.

PR #133, opened 2018-06-02
Initial version of pfs-swift-load tool... only PUT case supported

edmc-ss:


edmc-ss:
I don't intend on merging this into development until the aforementioned remaining options are implemented... those shouldn't take too long fortunately now that the framework is in place and the PUT case is already complete (save for sending non-zero Content-Length of course).

PR #135, opened 2018-06-05
Update the version of Swift to test against

tipabu:


tipabu:
This has been obviated by an even newer version of Swift.

PR #136, opened 2018-06-05
pfs_middleware: log RPC durations

smerritt:
HTTP requests can be slow because Swift was slow or because proxyfsd
was slow. By adding some timing info (at DEBUG level), we should be
able to tell which thing is slow.

This is pretty straightforward, but I haven't got a working development environment to try it in. Unit tests pass, though.


tipabu (approved):
Yeah, seems sane. Roughly how many of these would you expect to see in a second, or a minute? Regardless, since it's at DEBUG, shouldn't matter *too* much...

edmc-ss:
For the typical ProxyFS RPCs, I'd say they are:
  1 RPC per HEAD (stat)
  2 RPCs per GET (stat & get-read-plan)
  2 RPCs per PUT (provision-object & put-complete)
Do I have this about right @smerritt ?

tipabu:
This has languished plenty long enough...

PR #146, opened 2018-07-09
Add NoAuthIPAddr option for go swiftclient

bloodeagle40234:
Previously swiftclient go binding written in ProxyFS repository is pinning the no auth proxy server ip address as 127.0.0.1 but that hard coded configuration gives us constraints like 1. no-auth proxy sever cannot be listening any internal ip address for any no-auth fuse system. 2. proxyfsd cannot be configured to look up existing proxy-server for testing. And also the deployment model that proxyfsd should be located swift proxy would give more operational consideration because backend swift requires Ring distribution for all swift proxy nodes in the cluster.

This new option called NoAuthIPAddr is for setting the IP address from proxyfsd to no-auth swift proxy server enables us to separate the deployment operation between proxyfsd and backend swift and make the flexible configuration for us in my opinion.


edmc-ss:
In swiftclient/config.go, the change made immediately mandates that the config file specifies the SwiftClient.NoAuthIPAddr. This is definitely my preference i.e. that there be no defaults in the code. That said, the practice of providing a temporary default until such time that scripts/tools building the ProxyFS configs catch up might be nice.

If you don't mind, I'd like to make that temporary default present and merge the result. Ok?

swiftclient/config.go
```diff
@@ -76,7 +77,12 @@ func Up(confMap conf.ConfMap) (err error) {
 		return
 	}
 
-	globals.noAuthStringAddr = "127.0.0.1:" + strconv.Itoa(int(noAuthTCPPort))
+	noAuthIPAddr, err = confMap.FetchOptionValueString("SwiftClient", "NoAuthIPAddr")
+	if nil != err {
+		return
+	}
+	globals.noAuthStringAddr = noAuthIPAddr + ":" + strconv.Itoa(int(noAuthTCPPort))
```
edmc-ss, line 84: This is where my comment about providing a temporary default for NoAuthIPAddr would go... if you are ok with it...

bloodeagle40234:
Temporary default sounds ok to me to migrate until the option is available as required.

edmc-ss:
Ok - great! I'll pull this I to 1.6 we hope to tag this Friday. Thanks for the PR.

edmc-ss:
Brought over PR into swiftstack/ProxyFS repo as PR https://github.com/swiftstack/ProxyFS/pull/155

PR #147, opened 2018-07-10
Updated fs & httpserver to expose ReadPlan's on file inodes

edmc-ss:


edmc-ss:
@gerardgine - Please excuse the branch name... it was originally named for the first step I did to enable reporting of the ExtentMap's entire contents... but then, well, you were so fast getting me the HTML/JS that I just went ahead and finished everything :-).

As you mentioned in Slack, I believe you wanted to unify some layout discrepancy you noted on the SnapShot page... so here's the PR we can amend with that change if you want...

edmc-ss:
Added httpserver.gitDescribeOutput string... now we need to communicate it out of the HTTP Server both on the web page as well as a normal GET response...

edmc-ss:
OK... so we have a problem...
"git" is apparently not in $PATH when running the CI build :-(

edmc-ss:
Resolution: Travis is doing a git clone --depth=50... and git describe just happens to not find any tags looking back 50 commits... indeed, it will find 1.5.3 84 commits back. If --depth had been at least 84, it wouldn't have failed...

edmc-ss:
On to @gerardgine for his updates to the httpserver/html_templates.go to display the output of `GET /version` in the appropriate way on each of the web pages. We should be good to merge to development after that.

PR #148, opened 2018-07-11
Discard inode cache

bschatz-swift:
Added a timer thread that discards inodes from the inode cache when the inode cache size is above a certain number of bytes.

The code works when testing with proxyfsd.   However, the FS test hangs.

While I am debugging that - I would appreciate code review feedback.  I have targeted some ares with 'TODO' since I have open questions.


fs/treewalk.go
```diff
@@ -1241,7 +1241,7 @@ func (sVS *scrubVolumeStruct) scrubVolumeInode(inodeNumber uint64) {
 
 	defer sVS.jobReleaseParallelism()
 
-	inodeLock, err = sVS.volume.initInodeLock(inode.InodeNumber(inodeNumber), nil)
+	inodeLock, err = inode.InitInodeLock(sVS.jobStruct.volumeName, inode.InodeNumber(inodeNumber), nil)
```
edmc-ss: I guess this surprised me... I thought you had said you moved this API to DLM... and not INODE.

As for adding APIs to package Inode, I'm quite a bit more critical... as all of the existing APIs in package Inode are methodically tested... so I'll have a high bar for agreeing that an API needs to be added. Let's see the test code for this...

inode/config.go
```diff
@@ -94,6 +98,7 @@ type globalsStruct struct {
 	corruptionDetectedFalseBuf         []byte                        // holds serialized CorruptionDetected == false
 	versionV1Buf                       []byte                        // holds serialized Version            == V1
 	inodeRecDefaultPreambleBuf         []byte                        // holds concatenated corruptionDetectedFalseBuf & versionV1Buf
+	inodeSize                          uint64                        // size of in-memory inode struct
```
edmc-ss, line 101: As we've discussed... I do not agree with this number as it is actually only capturing the relatively small "constant" part of the size of an inode. I realize your and @charmster's differing opinions on this that "users think in terms of size, not count", but that is very, very hard to quantify. Sure, we could do it... but I doubt that's worth the effort to be honest. So I'd much prefer sticking with counts.

To that end, there is tremendous value in following a convention set by many other such mechanisms. All of those use counts... so it is bucking the pseudo standard as well to switch to size for just this one. It's kind of like somebody saying "hey, I like 6 character tab stops instead of the 4 character tab stops... so in the code I write I'll use 6". Does this make sense?

inode/config.go
```diff
@@ -116,6 +121,8 @@ func Up(confMap conf.ConfMap) (err error) {
 		flowControl                                    *flowControlStruct
 		flowControlName                                string
 		flowControlSectionName                         string
+		LRUCacheMaxBytes                               uint64
+		LRUDiscardTimeInterval                         time.Duration
```
edmc-ss: As we've discussed, I dislike any sort of polling and would much prefer triggering the code that evicts at the point of insertion. Sure, it could be done in the background (by simply launching a goroutine)... indeed, that is the pattern for lots of things going on in ProxyFS (check out the quota limiting the number of chunked PUTs underway). Sure, it's a bit complicated to manage, but is far more deterministic than polling which could fall behind and is, thus, difficult to tune.

What I'm suggesting is we have a (lock-protected) flag that says whether a eviction goroutine is already running or not such that when inserting in the inodeCache would trigger an eviction that a new goroutine is only launched when that flag is false. Hope this makes sense.

inode/config.go
```diff
@@ -192,6 +199,9 @@ func Up(confMap conf.ConfMap) (err error) {
 	globals.accountMap = make(map[string]*volumeStruct)
 	globals.flowControlMap = make(map[string]*flowControlStruct)
 
+	var tmpInode inMemoryInodeStruct
+	globals.inodeSize = uint64(unsafe.Sizeof(tmpInode))
```
edmc-ss, line 237: See my earlier comment.

If you look at the inMemoryInodeStruct, you will see all manner of by-reference fields.
So I guess I shouldn't be surprised that the value calculated here is 300 Bytes...
I believe the actual average size is more like 1KB to 2KB of course depending on how
many extents are in a fileInode's ExtentMap and how many elements are in a dirInode's DirMap.

inode/config.go
```diff
@@ -210,6 +220,28 @@ func Up(confMap conf.ConfMap) (err error) {
...
+		if nil != err {
+			return
+		}
+		volume.inodeCacheLRUTickerInterval = LRUDiscardTimeInterval
+		volume.inodeCacheLRUTicker = time.NewTicker(volume.inodeCacheLRUTickerInterval)
+
+		go func() {
+			for range volume.inodeCacheLRUTicker.C {
+
+				// TODO - Do we have to worry about the volume being removed?
```
edmc-ss: Indeed - is there no sync.WaitGroup that the volume deletion code can at least Wait() on?

inode/config.go
```diff
@@ -894,7 +926,17 @@ func ExpandAndResume(confMap conf.ConfMap) (err error) {
 }
 
 func Down() (err error) {
+
+	// Stop the goroutine which culls the inodeCache
+	// TODO - Any issues with the pause/resume code?????
```
edmc-ss: I think inodeCache eviction is questionable to run while paused... our only requirement is that inode not call any layer below it while paused. Cache eviction does involve taking out a DLM Lock, right? That said, I'm not sure what the DLM layer expects - so it may just happen to work - but this is, I think, a totally unnecessary violation of the Pause-semantic.

I'd just do the same thing during Pause (for EVERY volume) that we need to do when deleting a volume - i.e. "drain" any inodeCache eviction goroutine(s). Given the Pause semantics, we will also be assured that no "outside" stimulus will cause us to need to find room in the inodeCache, right?

To be clear - if we are to continue this "polling" approach, we not only need to drain any ongoing eviction goroutine logic but we also need to pause/halt the polling, ok?

inode/config.go
```diff
@@ -894,7 +926,17 @@ func ExpandAndResume(confMap conf.ConfMap) (err error) {
...
+	for _, volume := range globals.volumeMap {
+		if volume.active {
+			volume.inodeCacheLRUTicker.Stop()
```
edmc-ss: Excellent! Just do this in Pause as well... plus remember to restart it for all active volumes at time of Resume/Expand.

inode/inode.go
```diff
@@ -351,13 +352,78 @@ func (vS *volumeStruct) inodeCacheTouch(inode *inMemoryInodeStruct) {
...
 
+// TODO: What happens if the volume is removed?  What if
+// failover?
+func (vS *volumeStruct) inodeCacheDiscard() {
+	inodesToDrop := uint64(0)
+
+	vS.Lock()
+
+	if (vS.inodeCacheLRUItems * globals.inodeSize) > vS.inodeCacheLRUMaxBytes {
+		// Check, at most, 1.25 * (minimum_number_to_drop)
```
edmc-ss, line 362: Unless you always evict the total # of inodes in excess of the limit at each timer tick, it is quite possible that a too large timer tick interval may fall further and further behind... hence one of a few reasons why I prefer a "triggered-by-exceeding-the-limit" mechanism over a "polling" one.

inode/inode.go
```diff
@@ -351,13 +352,78 @@ func (vS *volumeStruct) inodeCacheTouch(inode *inMemoryInodeStruct) {
...
+			inodeRWLock.Unlock()
+			vS.inodeCacheTouchWhileLocked(ic)
+			continue
+		}
+
+		vS.Unlock()
+
+		var ok bool
+
+		ok, err = vS.inodeCacheDrop(ic)
```
edmc-ss, line 396: This, I think, is awkward... you could simply drop ic while still locked... that way, you'll always be dropping the head element. That said, since ...Drop() can drop out of the middle of the LRU, this is actually fine.

inode/inode.go
```diff
@@ -351,13 +352,78 @@ func (vS *volumeStruct) inodeCacheTouch(inode *inMemoryInodeStruct) {
...
+		if err != nil || !ok {
+			pStr := fmt.Errorf("The inodes was not found in the inode cache - ok: %v err: %v", ok, err)
+			panic(pStr)
+		}
+
+		// TODO - I am releasing the locks out of order here.   That seems wrong.
+		// We acquire the locks in this order:
+		// 1. Volume lock
+		// 2. DLM lock for inode
```
edmc-ss: I think the reason it is ok to do (1) before (2) is because you are doing a TryLock for (2).
So I don't think it's actually wrong... Does this make sense?

inode/inode.go
```diff
@@ -351,13 +352,78 @@ func (vS *volumeStruct) inodeCacheTouch(inode *inMemoryInodeStruct) {
...
+}
+
 func (vS *volumeStruct) inodeCacheDropWhileLocked(inode *inMemoryInodeStruct) (ok bool, err error) {
 	ok, err = vS.inodeCache.DeleteByKey(inode.InodeNumber)
 	if (nil != err) || !ok {
 		return
 	}
 
-	// TODO: Remove inode from inodeCacheLRU
+	// QUESTION - Is the above TODO still correct?????
```
edmc-ss: Yeah - I must have forgotten to remove this TODO - since I implemented it.

BTW, personal preference... I would have much preferred you actually "called" the routines I wrote (as you would have had to do had you converted to list.List, right?)... I really don't see why you needed to "embed" my clean/simple func inside your more complex function. At least that's the way it appears in the GitHub/Gerrit-like review tool. Why didn't you just call my function?

inode/locker.go
```diff
@@ -0,0 +1,95 @@
+package inode
```
edmc-ss, line 1: Ok... so to follow the consistency of package Inode, I would have expected to see these APIs listed in inode/api.go. Further, rather than passing the volumeName as an argument, these should have been "methods" of the inode.volumeStruct, right? Sure, package fs needs to call them too I suppose... but in every case it does so, it has the inode VolumeHandle, right?

inode/locker_test.go
```diff
@@ -1,27 +1,32 @@
...
+	testVolumeHandle, err := FetchVolumeHandle("TestVolume")
+	if nil != err {
+		t.Fatalf("FetchVolumeHandle(\"TestVolume\") failed: %v", err)
+	}
+
+	volume := testVolumeHandle.(*volumeStruct)
 
 	// Test: get a lock for vol A inode 2
-	srcDirLock, err := mS.volStruct.initInodeLock(inodeNumber, nil)
+	srcDirLock, err := InitInodeLock(volume.volumeName, inodeNumber, nil)
```
edmc-ss: Yeah, here is the first example... I think fs's volume should already have an inode VolumeHandle, right?

Anyway, I like that the test code moved, too :-).

But where are all the changes to fs's calls to DLM?
I would have expected a ton of them...

edmc-ss (commented):
So I have a lot of comments... sorry.
Many are my personal preferences... hope you don't mind too much.

edmc-ss:
Apologies... I was confused by not seeing the changes in fs/api_internal.go... these are the calls to the DLM wrappers that have now moved to Inode so not seeing them was confusing.

fs/api_internal.go
```diff
@@ -145,9 +145,9 @@ func (vS *volumeStruct) inFlightFileInodeDataFlusher(inodeNumber inode.InodeNumb
 
 	// Act as if a package fs client called Flush()...
 
-	inodeLock, err = vS.initInodeLock(inodeNumber, nil)
+	inodeLock, err = inode.InitInodeLock(vS.volumeName, inodeNumber, nil)
```
edmc-ss: Ok... now that I could see the formerly collapsed diff of fs/api_internal.go, I'll codify my preference here... Instead of ```inode.InitInodeLock(vs.volumeName, inodeNumber, nil)```, I would have expected to see ```vS.inodeVolumeHandle.InitInodeLock(inodeNumber, nil)```. See the difference?

Meanwhile, that would mean I expected to see the prototype for inode.InitInodeLock (and friends) to instead of being ```func InitInodeLock(volumeName string, inodeNumber InodeNumber, callerID dlm.CallerID)``` to be ```func (vS *volumeStruct) InitInodeLock(inodeNumber InodeNumber, callerID dlm.CallerID)```.

Of course, as I think I've verbalized elsewhere, I see that you put these APIs in inode because they reference InodeNumber. I think something we all should internalize is that InodeNumber is a uint64... comes from the Nonce space shared by *all* enumerated resources that might conceivably be locked... and, hence, DLM could merely expose these APIs but calling what is now termed the ```inodeNumber InodeNumber``` argument as, simply, ```key uint64```.

I realize you and @charmster have, at times, preferred e.g. that the Nonce spaces for InodeNumbers be distinct from, e.g., LogSegmentNumbers... but that's simply not a trivial change...with very little benefit (if any). So again, let's all agree on the current implementation of ProxyFS having this property (my hope) and reconsider the choice to place these InodeNumber-specific wrappers in (fs previously but now) inode. Just my $0.02.

bschatz-swift: Added a receiver for the functions related to wrapping DLM locks for inodes.

inode/locker_test.go
bschatz-swift: Yes - changed

inode/locker.go
bschatz-swift: Done

inode/inode.go
bschatz-swift: Must be a github interface problem - I am only calling your functions unless I am missing something

inode/inode.go
bschatz-swift: Makes sense - wanted to call it out so that we are all on the same page

inode/inode.go
bschatz-swift: Let me think this one over some more - to be honest I think it is cleaner to keep the lock and call inodeCacheDropLocked()...

inode/config.go
bschatz-swift: done

inode/config.go
bschatz-swift: We now start the discard routine in Up() and ExpandAndResume().   We stop the discard routine in Down() and PauseAndContract()

inode/config.go
bschatz-swift: Not that I am aware of...

fs/treewalk.go
bschatz-swift: Fix per comments below...

Issue #149, opened 2018-07-11
Support delimiter requests for container listings


tipabu:
Currently, trying to list a ProxyFS "container" through the `pfs` middleware ignores `delimiter` query params, causing all objects in the container to be listed.

This would probably be fairly cheap to do for `delimiter=/`, but the more general case still requires that we walk the whole tree. Maybe it'd be enough for 90% of use cases to support `/` and 400 any other delimiters?

Fixing this should solve several Swift and Swift3 functional tests (as well as make large volumes manageable for Swift API users).

gerardgine:
Changes included in tag 1.5.3.1 and merged into development + stable branches.

PR #157, opened 2018-07-18
Added delimiter support

gerardgine:


jrpcfs/middleware_test.go
```diff
@@ -697,6 +697,44 @@ func TestRpcGetContainerPrefixAndMarker(t *testing.T) {
...
+	}
+	response = GetContainerReply{}
+	err = server.RpcGetContainer(&request, &response)
+
+	assert.Nil(err)
+	assert.Equal(1, len(response.ContainerEntries))
+	ents = response.ContainerEntries
+	assert.Equal(".git/logs/refs", ents[0].Basename)
+}
+
```
edmc-ss, line 742: Interesting... hadn't noticed this before - and please don't feel like you need to change... but the convention in Go test code is to reference your `t *testing.T` argument... and invoke APIs relative to that (see the description of what's available for `type T` at https://golang.org/pkg/testing/).

I take it all these `assert.*()` calls will simply panic. `go test` will definitely see this as a failed test...but in a less clean way I suppose. Alas, package test didn't deign to include all these assert functions you (and Sam) are using from package github.com/stretchr/testify/assert :-(.

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -1200,8 +1205,16 @@ def _json_container_get_response(self, container_entries, account_name):
...
+            if delimiter != "" and "IsDir" in ent and ent["IsDir"]:
+                json_entries.append({"subdir": ent["Basename"] + delimiter})
+
         return json.dumps(json_entries)
 
+    # TODO: This method is usually non reachable, because at some point in the
+    # pipeline, we convert JSON to XML. We should either remove this or update
+    # it to support delimiters in case it's really needed.
+    # Same thing probably applies to plain text responses.
     def _xml_container_get_response(self, container_entries, account_name,
```
edmc-ss, line 1218: Doesn't this get called if the request specifies XML response?

edmc-ss (approved):
Looks great!

Plus I don't anticipate any merge conflicts with development versus your hotfix on top of 1.5.3 (he says hopefully).

jrpcfs/middleware_test.go
bschatz-swift: @edmc-ss The assert calls will show up as a failed test at the end of the run.  They will not panic.

bschatz-swift: @edmc-ss The assert() calls don't panic.  They just fail the individual test.
After all tests are executed a summary of failures is printed.

bschatz-swift (approved):
I only looked at the Go related changes

pfs_middleware/pfs_middleware/middleware.py
timuralp: Depends on the version of Swift -- unfortunately. With Swift 2.16.0 or later (and 2.15.3 SwiftStack Swift release), this middleware will only ever see content type as application/json. There is a different piece of middleware -- `listing_formats`, which requests JSON back and then mutates it to any form that the client requests (plain or xml). We can probably remove this additional formatting code, actually, and always return JSON if we have a guarantee about the Swift version.

jrpcfs/middleware_test.go
edmc-ss: Aha! That is because of line 702: assert := assert.New(t)

That line "shadows" the package (really a subdir of a package) named assert with an instance of the (sub)package's wrapper around a *testing.T... makes sense.

edmc-ss:
Can we be sure to comment on and close out Issue https://github.com/swiftstack/ProxyFS/issues/149

jrpcfs/middleware_test.go
gerardgine: To be honest, I just copied and pasted one of Sam's tests and adapted it to my additions. What I know is I ran the tests, and they passed. In this case, yes, thanks to the line @edmc-ss pointed out.

gerardgine:
Closing this PR, as the changes were finally applied to hotfix 1.5.3.1 and merged into development and stable.

PR #158, opened 2018-07-18
Adjust tunables, fix issues with start/stop timers, what entries removed, etc

bschatz-swift:
Adjust tunables, start/stop of discard timer and when to discard

    Adjusted the tunables based on more testing.

    Fix issues related to when we start/stop the timers

    Fix bug where we were incorrectly discarding inodes from the inode cache.

    Moved execution of for() loop under if statement since no need to test for() loop
    if if statement did not pass.

### DIFF
--- a/inode/config.go
+++ b/inode/config.go
@@ -121,33 +121,36 @@ func startInodeCacheDiscard(confMap conf.ConfMap, volume *volumeStruct, volumeSe
 	)
 	LRUCacheMaxBytes, err = confMap.FetchOptionValueUint64(volumeSectionName, "MaxBytesInodeCache")
 	if nil != err {
-		return
+		LRUCacheMaxBytes = 10485760 // TODO - Remove setting a default value
+		err = nil
 	}
 	volume.inodeCacheLRUMaxBytes = LRUCacheMaxBytes
 
 	LRUDiscardTimeInterval, err = confMap.FetchOptionValueDuration(volumeSectionName, "InodeCacheEvictInterval")
 	if nil != err {
-		return
+		LRUDiscardTimeInterval = 1 * time.Second // TODO - Remove setting a default value
+		err = nil
 	}
 
-	if LRUDiscardTimeInterval == 0 {
+	if LRUDiscardTimeInterval != 0 {
+		volume.inodeCacheLRUTickerInterval = LRUDiscardTimeInterval
+		volume.inodeCacheLRUTicker = time.NewTicker(volume.inodeCacheLRUTickerInterval)
+
+		logger.Infof("Inode cache discard ticker for 'volume: %v' is: %v MaxBytesInodeCache: %v",
+			volume.volumeName, volume.inodeCacheLRUTickerInterval, volume.inodeCacheLRUMaxBytes)
+
+		// Start ticker for inode cache discard thread
+		go func() {
+			for range volume.inodeCacheLRUTicker.C {
+
+				volume.inodeCacheDiscard()
+			}
+		}()
+	} else {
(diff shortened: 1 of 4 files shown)


inode/config.go
```diff
@@ -121,33 +121,36 @@ func startInodeCacheDiscard(confMap conf.ConfMap, volume *volumeStruct, volumeSe
...
 
 	LRUDiscardTimeInterval, err = confMap.FetchOptionValueDuration(volumeSectionName, "InodeCacheEvictInterval")
 	if nil != err {
-		return
+		LRUDiscardTimeInterval = 1 * time.Second // TODO - Remove setting a default value
+		err = nil
 	}
 
-	if LRUDiscardTimeInterval == 0 {
+	if LRUDiscardTimeInterval != 0 {
```
edmc-ss, line 135: I guess we technically shouldn't assume 0s is == 0... time.Duration just happens to be an int64 and I gather the parser is being friendly and not requiring us to type cast 0 to be either int64(0) or, more completely, time.Duration(0).

inode/inode.go
```diff
@@ -373,55 +373,52 @@ func (vS *volumeStruct) inodeCacheDiscard() {
 		inodesToDrop = (vS.inodeCacheLRUItems * globals.inodeSize) - vS.inodeCacheLRUMaxBytes
 		inodesToDrop = inodesToDrop / globals.inodeSize
 		inodesToDrop += inodesToDrop / 4
-	}
-	for ((vS.inodeCacheLRUItems * globals.inodeSize) > vS.inodeCacheLRUMaxBytes) &&
-		(inodesToDrop > 0) {
-		inodesToDrop--
+		for ((vS.inodeCacheLRUItems * globals.inodeSize) > vS.inodeCacheLRUMaxBytes) &&
```
edmc-ss: I now accept what @bschatz-swift & @charmster are trying to say here that "users" don't really care about # of inodes but rather space... but I guess I see no advantage in internally calculating size everytime... especially since it is an approximation anyway.

Couldn't we just "compute" vS.inodeCacheLRUMaxItems by taking vS.InodeCacheLRUMaxBytes and dividing by globals.inodeSize at volume creation time?

inode/inode.go
```diff
@@ -373,55 +373,52 @@ func (vS *volumeStruct) inodeCacheDiscard() {
...
-		//
-		// We then release the volume lock before deleting the inode from the cache and
-		// then releasing the DLM lock.
-		inodeRWLock.Unlock()
-
-		// vS.inodeCacheDrop() removed the inode from the freelist so
-		// the head is now different
-		vS.Lock()
 	}
 	vS.Unlock()
```
edmc-ss: So, just to be clear, we will hold the vS.Lock while in the loop?

If that is so, then I think the test at the top of the loop can be simplified to just `inodesToDrop > 0` since it can only be "this loop" that decrements vS.inodeCacheLRUItems, right? If two such timer tick instances happen to be running at the same time, only one of them will have the vS.Lock and the other will be blocked... and presumably find nothing to do when they eventually get the lock.

edmc-ss (approved):
This looks great - I only had a couple of stylistic comments.

Re your changing of the two .conf files and lamenting feeling the need to go back and match those values in the half dozen or so *_test.go files, I don't think you should feel the need to change them. There is no requirement that they match.

Indeed, I've on occasion been optimizing our test cases to e.g. have fewer Swift retries...

inode/inode.go
bschatz-swift: We do hold the volume lock the whole time.   I will change the for loop since it would be clearer.

bschatz-swift:
I updated the for loop.  I am now rebuilding so that we can start our long running test.

PR #167, opened 2018-08-02
Add LRU of inode.inFlightLogSegmentStruct's

edmc-ss:
This will be used to re-implement the enforcement of the MaxFlushTime
logic (currently up in fs) as well as a new "lighter-weight" mini-flush
mechanism (triggered by the MaxFlushTime expiration as well as the
ChunkedPutContext quota exhaustion/starvation upcall).


inode/file_flusher.go
```diff
@@ -346,30 +443,58 @@ func (vS *volumeStruct) doFileInodeDataFlush(fileInode *inMemoryInodeStruct) (er
...
-		inFlightLogSegment.fileInode.Unlock()
+		openLogSegmentLRURemove(inFlightLogSegment)
+
+		// And we are done
 		inFlightLogSegment.fileInode.Done()
+	}
+}
+
+func chunkedPutConnectionPoolStarvationCallback() {
+	var (
```
bschatz-swift: This is only going to flush 1 inode at a time.   Do we want it to flush multiple inodes until we get below a certain limit or am I missing something?

charmster:
Please do not close this as complete with these changes.  These fixes are probably an improvement over the current code, but are not good enough for a production release.

(these comments are not complete.  i hope to write a more extensive commentary tomorrow.  meanwhile:)

1) the `(*ChunkedPutContext) Active() (bool)` should will provide inconsistent results and should not be used.  there are two problems here:
1.1) the "ChunkedPutContext" active state is not what you think it is.  it distinguishes between requests that have an outstanding `PUT` operation to the Swift storage and operations that are in the middle or a retry.  when the first attempt fails, the `active` state changes to `false` until we initiate a retry, at which point it becomes true.  you are using this state up in the inode layer in way that is not, _i think_, consistent with the usage in the swiftclient layer.
1.2) the "ChunkedPutContext" active state is (or should be) private to the swiftclient layer.  if higher level layers, like _inode_, want to know if a log segment is  _active_ they should implement their own mechanism that is consistent with the usage within that layer.

2) i don't see any mechanism that flushes "idle" chunked put connections.  we already know that the noauth proxy server will fail PUT connections that are idle for some amount of time, and that amount of time is, _apparently_, less then flush timeout.  we need a daemon that Close()s idle connections before the Swift timeout for PUT operations (whatever that delay might be, but 10 sec seems reasonable to me).

3) even after point 2) is taken into account, and we have a mechanism to flush PUT request that have been completely idle for N seconds, we also ned mechanism that Close()s PUT connections that have been open for more than R seconds (where N may not be the same as R).  closing PUT connections _after_ we have run out of them, so the requestor that wants a new PUT connection must block, is far too late and inappropriate.  we should be flushing proactively.

4) i have some other, more picayune, criticisms that i will try to write down.

inode/file_flusher.go
edmc-ss: Given the synchronous nature of the call... I see no real benefit in holding up the allocation any longer than necessary. There is no longer any so-called "starvationMode" we transition in and out of that might be expensive in some respects... so I deleted the concept of looping. One less thing to have difficulty tuning, right?

edmc-ss:
The meaning of chunkedPutContextStruct.active (what swiftclient.Active() reports) was strictly a "did we order these things correctly" aid - I hope we never actually call it :-)

What chunkedPutContextStruct.active is *supposed* to mean is, if true, we've started the Chunked PUT... and are able to now send a chunk. If one were to Close() the context... we have now sent the zero-length chunk telling Swift the PUT data is complete... please respond. At this point, we cannot send another chunk.

So it's merely a correctness check. If clients (and our retry logic) are coded correctly, the test should never surprise us, right :-).

edmc-ss:
The MaxFlushTime check is actually up in package fs (where it was before I moved the open chunked PUT connection LRU/flusher part of it). In fs, the MaxFlushTime is enforced across *all* chunkedPutContext's associated with the FileInode... not just one.

To see that in action, please have a look at the inFlightFileInodeDataStruct definition in fs/config.go. These are on the globals.inFlightFileInodeDataList... monitored by inFlightFileInodeDataTracker() (fs/api_internal.go). The logic is a bit convoluted to handle the various conditions under which we would want to flush (admittedly now simpler since the starvationCallback() is now down in inode... so perhaps this could be simplified) - but I do believe the MaxFlushTime is enforced effectively. Were it not, we'd now starve Chunked PUT Contexts and never come out of it with the way this PR/branch now does things (and before it for that matter).

edmc-ss:
I think I've address @charmster 's Points 1 & 2... not sure what 3 is asking...
Eagerly awaiting more details on 4 :-)

edmc-ss:
Ah - I think I get 3... if we were to create a chunked PUT connection but not send anything to it, I guess that we would have a problem if we only timed out cases where we'd actually sent data to it. Turns out we don't have that fine a logic. Indeed, the way it works today, we could have a 10 second timer running even though all data has been flushed and there are no active chunked PUT connections.

This works this way because, at the fs layer, he tracks FileInodes that may possibly have one or more chunked PUT connections (and in memory "dirty" metadata as well)... but it doesn't really know that it does. For instance, if I have a MaxFlushSize of 1MB... and do an fs.Write() to a previously unwritten-to FileInode with a 1MB buffer, inode layer will Fetch a chunked PUT context, send the 1MB to it, and immediately notice that "hey, I just hit MaxFlushSize" and launch a goroutine to Close() it. When inode returns to fs.Write(), it may be the case the Close() has, in fact, already completed... in which case fs continues to track it (as needing a MaxFlushTime-limited Flush) even though it doesn't.

Hope this addresses @charmster's (3).

Meanwhile, I had to look up "picayune" :-)

inode/file_flusher.go
bschatz-swift: Makes sense

edmc-ss:
So the way optional/variadic args work in Go is that the last arg can be a slice of a single type... hence the length of the slice is how many you supply. Sure enough you can supply an empty slice (len == 0), but if you don't supply the optional/variadic arg at all, you are actually passing nil. Hence, the nested comparison first against nil and then against a length of 1 (since we only allow a single optional VolumeName argument.

Does this make sense? I suppose another variant would be to remove the variadic nature of the argument and make it a singular string... and tell all callers not wanting to use the lone connection reserved for the volume to pass either nil or "". But this was the easiest way to make the change without such a global change :-).

swiftclient/api.go
```diff
@@ -109,8 +127,8 @@ func ObjectDelete(accountName string, containerName string, objectName string, o
 }
 
 // ObjectFetchChunkedPutContext provisions a context to use for an HTTP PUT using "chunked" Transfer-Encoding on the named Swift Object.
-func ObjectFetchChunkedPutContext(accountName string, containerName string, objectName string) (chunkedPutContext ChunkedPutContext, err error) {
-	return objectFetchChunkedPutContextWithRetry(accountName, containerName, objectName)
+func ObjectFetchChunkedPutContext(accountName string, containerName string, objectName string, useReserveForVolumeName ...string) (chunkedPutContext ChunkedPutContext, err error) {
+	return objectFetchChunkedPutContextWithRetry(accountName, containerName, objectName, useReserveForVolumeName...)
 }
```
charmster: There's no good reason to make this a variadic function.  If we need to add a "volume name" parameter then just add an additional paramater and be done with it.

ObjectFetchChunkedPutContext() is only called in 19 places.  Updating the 18 of them not currently affected to add ", nil" to the end of the function call would take less time then we're probably going to spend arguing about this, particularly since the compiler will flag any places we missed.

Assuming we go forward with this change, I'd be happy to go through and add the nil argument to the other 18 places and strip out code for the variadic arguments.

charmster (changes requested):
I've been thinking about this a lot and _I'm not sure that we should do anything to address the "starvation issue"_ and I'm quite sure that we shouldn't make the elaborate and complicated set of changes suggested here.

I think that, at most, we should do something to Close idle chunked put connections faster (perhaps after they are idle for 1 sec), and perhaps reduce that delay a bit (to 100 msec?) if we run out of chunked put connections.  _Doing anything more aggressive is probably going to lead to thrashing_, where we actually hurt our throughput and probably hurt the _average_ latency as well.

The reason is simple: if we're actually out of chunked put connections (i think the pool size is 1024 on sm-p1) then that means (or should mean) that there are more than 1024 files actively being written to.  If the connections are _not idle_ then closing them in between writes means creating more objects and paying the object create time/object close time penalty more often.  Its pretty easy to see if that if there are 1024 connections and 2048 active writes, aggressively closing the connections will probably result in ~1 write per object, and a hell of lot of time spent creating and closing objects, which is time not spent writing actual data.

-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

But before arguing about that, I'd like to take two steps back and consider how we got to this solution ...

I believe we were running tests on sm-p1 and saw a hang where all 1024 chunked put connections were in use and headhunter go routine that were trying to do a checkpoint and stuck because they couldn't get a chunked put connection.  It looked like a deadlock (and it may have been an actual deadlock; I'm not sure).

So we added a "starvation callback" routine that the swiftclient code could call if it ran out of chunked put connections.  The starvataion callback routine was supposed to free up some chunked put connections so we could make progress.

Unfortunately, the first attempt at writing a starvation callback routine didn't work out; it became obvious that the starvation callback was involved in a deadlock since it tried to acquire the _inode lock_ and other code acquired the _inode lock_ and then tried to get a chunked put connection (since chunked put connections come from a limited pool it is effectively a lock).  Its at this point that this Pull Request and an alternate Pull Request that i wrote diverge.

But (taking the same metaphorical step back), we still haven't figured out why "starvation" occurred in the first place.  There is a mechanism at the fs layer that is supposed to Close chunked put connections after 10 sec (regardless of whether they are idle or not).  If that worked as expected we would never see a hang.  At most we would see a number of go routines waiting to get a chunked put connection, but those routines would get one after some number of seconds and continue to make progress (the number of seconds would depend on how heavily over-subscribed the chunked put connection pool is).  Note: in order to get into this situation, we would need to have more than 1,000 simultaneous write requests (plus some slop for checkpoints that are in progress), which none of our tests do.

So, I suspect we have a yet undiagnosed deadlock in the fs layer file flushing.

-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

Getting back to this PR, there's a bunch of changes that I think needlessly complicate the code.  Code complexity makes code less reliable and harder to maintain, and is therefore to be avoided.  I think we want the minimum amount of code sufficient to solve the problem.

I also suspect that the code in this Pull Request still contains a deadlock.  The fact that we haven't hit a deadlock in testing doesn't mean it doesn't exist, it just means we haven't hit it.  The fact that I can't spell out exactly how the deadlock would occur also doesn't mean it doesn't exist.

If the code were slightly different, I can point to one potential deadlock:

_This code allows an unlimited number of chunked put connections to be used for checkpoints_ (ignoring any limit on the number of chunked put connections).  _If that were not the case_, the following deadlock probably exists:

o Assume the inode starvation callback function, `inode.chunkedPutConnectionPoolStarvationCallback()` is called because we are out of connections.  It will call inode.inFlightLogSegmentFlusher(), which will get the lock on the inode.

o Assume the inode lock is held by another goroutine, which is calling inode.flushInode(), which is trying to serialize the inode into the inode tree.  It attempts to get the headhunter lock (volume lock) and is blocked on it.

o Assume the headhunter lock is held by a goroutine trying to do a checkpoint.  That routine needs a new chunked put conn
ection, so it calls into `swiftclient.ObjectFetchChunkedPutContext()` to get one, which calls `swiftclient.acquireChunkedConnection()`.

o Assume that the `swiftclient.acquireChunkedConnection()` actually obeyed the limits on the number of connections and invoked the starvation callback routine.  Then we would have deadlock.

Alternatively, if the code reserved a single connection per volume then it could wait for that connection to become available and make progress, at the cost of additional synchronization between `swiftclient.acquireChunkedConnection()` and `swiftclient.releaseChunkedConnection()`.  For that matter, acquireChunkedConnetion() shouldbe able to wait for _any_ connection to become available, _which should happen regardless of whether the starvation callback is invoked_ and we would make progress.

-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

I can see a couple _potential_ problems in `swiftclient.acquireChunkedConnection()`/`swiftclient.releaseChunkedConnection()` around the reserved connections for volume checkpoints.

the first is that is that if a checkpoint uses more than one chunked put connection concurrently, we will _leak_ the additional connections when they are released.  that's because swiftclient.releaseChunkedConnection() simply overwrites the slot for the volume's "reserved" connection:
```
                if keepAlive &&
                        (connection.connectionNonce == globals.connectionNonce) {
                        // Re-insert connection in globals.reservedChunkedConnection map

                        globals.reservedChunkedConnectionMutex.Lock()
                        globals.reservedChunkedConnection[connection.reserveForVolumeName] = connection
                        globals.reservedChunkedConnectionMutex.Unlock()
                }
```
Now, that's not a problem with the current implementation of headhunter checkpoints, but it will lead to subtle bug if and when we improve the concurrency.  Its quite a brittle interface between two completely separate pacakges.

Similarly, the reserved connection for a volume, once opened, is never closed.  Not only are we exceeding the "maximum" number of chunked put connections, there's not mechinism to get us down below this maximum.

We really shouldn't need these reserved connections in the first place (except for the problems with locking hierarchy introduced by `inode.chunkedPutConnectionPoolStarvationCallback()`) so let's just fix that and not introduce this complexity.
 -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

I have a similar complaint with `swiftclient.acquireChunkedConnection()`/`swiftclient.releaseChunkedConnection()`.  The synchronization is just too complicated.  There are a lot of gyrations to try and insure "fairness" in handing out chunked put connections (so the longest waiting caller gets the next connection).  We don't try fairness anywhere else in ProxyFS so I'm not sure why we're doing it here.

The additional complexity makes it harder to analyze whether the synchronization between the two routines is correct.  A synchronization mechanism that doesn't insure fairness is sufficient.

-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

To summarize, I think we might be barking up the wrong tree here.

If the starvation callback mechanism is here to avoid deadlock then we're in deep trouble, particularly because it will only be triggered after we've accumulated enough outstanding requests to use up all available connections ... which may never happen in a hang!

If its a performance improvement (as I think it is) then it should be less aggresssive about closing potentially active connections.

Regardless, the changes should be as simple as possible to accomplish their goal.

swiftclient/api.go
edmc-ss: Happy to make this change to remove the need for the variadic... it'll also make the comparison for the option's existence more clear. @bschatz-swift earlier made a comment about "why are you taking the length of the arg?" before... not very intuitive to be sure. But, then, variadics in positional argument lists have always been, uh, a hack imho :-).

edmc-ss:
I'm going to try and break up responses/action-items coming out of @charmster's list of comments... thanks, again, for the excellent review.

First off, I want to explain the FileInode Mutex's purpose. Back in the dawn of time... or, rather, the first version of package inode, we had a write throughput problem. One cause was determined to be that a given FileInode would fill up one 10MB LogSegment... then block while the Close() finished on that one... then start a new chunked PUT. This meant that the unlucky write that caused the Close() had to wait for it. @balajisrao did some modeling/experimentation of just what this was costing us... and it was deemed to be significant but certainly not responsible for the bulk of the slow write performance. Nevertheless, we set out to make the Close() happen in the background.

At this point, I should point out that we never actually succeeded at that! Unbeknownst to all of us, the logic to Close() off the LogSegment just filled in the background likely blocked the very next Write() call. We were only very marginally better than the prior situation where the last Write() to the LogSegment being Close()'d had to block until the Close() completed. But, hey, it wasn't thought to be the bulk of the bottleneck, so I guess we could be forgiven for not noticing that sooner.

The fix for this is in the PR... and basically says "hey, don't hold the FileInode Mutex while you are Close()-ing dummy! With this change, and the preexisting logic to handle any number of such being-Close()'d LogSegments, we are good.

This is why I was so un-supportive of trying to avoid obtaining the FileInode Mutex when checking for an existing "open" LogSegment... being as how that was the entire purpose of the FileInode Mutex in the first place. It was just implemented poorly. With the current corrections to: 1) Not hold the FileInode Mutex while "opening" a chunked Put context, and 2) Not holding the FileInode Mutex while (in the background) awaiting the Close() completion, we finally have implemented what I believe we set out to do so long ago. Indeed, with this proper implementation, it is literally no worse than the alternative proposed in the adjacent PR in that, surely, an identical Mutex needed to be grabbed while accessing the shared map[InodeNumber]*openPutContext... indeed, the shared map approach requires a much more global Mutex... not just one confined to the particular FileInode in question imho. So I much prefer this PR's approach for that reason alone.

edmc-ss:
I am really interested in understanding better the 100ms (or even smaller) LogSegment Close() timeout proposed by @charmster. It's an interesting thought. With the current MaxFlushTime (e.g. 10s), we are only really talking about "how far behind can the file system's metadata be for this FileInode being written to". That need not be the same as for the time we keep a LogSegment chunked PUT "open". Indeed, if the config sets MaxFlushTime > 60 seconds, this will likely cause 503 timeouts from the NoAuth Swift Proxy wondering why ProxyFS is taking so long to finish its chunked PUT :-).

In the very near term, we should update ProxyFS configurators to limit MaxFlushTime to much less than that 60 second timeout to avoid these 503s.

I propose we table that for the purposes of this PR and defer it to a later time.

edmc-ss:
Re deadlock hangs... I concur with @charmster that the starvation callback is really not "required" to avoid them... indeed, we didn't used to have them.

What was, however, sometimes causing the deadlock is actually known. We had a situation where MaxFlushTime (or the starvation callback in an earlier version of this PR) was attempting to complete a Flush() to "make room" for other users of the chunked PUT pool... when a checkpoint occurred. At that point, the checkpoint needs to get its own chunked PUT context... and it does so while holding the Volume Lock for headhunter. What this meant was that any metadata operation (including Flush()) might just be blocked attempting to perform said Flush(). It is, however, unclear why the Flush() code path would attempt to perform the headhunter updates before releasing the chunked PUT contexts.

This is why recognition that there is a justification - however avoidable in proper Flush() logic - for a reserved chunked PUT context allotment for the checkpoint operation. As luck would have it, we need only one (per Volume of course)... as, unlike for FileInode's, the checkpoint logic makes no attempt to "background" the waiting for the Close() completion before opening a fresh chunked PUT context. While such an optimization may help lower the checkpoint latency, that is not the current goal of this PR imho. And, recognizing that a checkpoint will include a "flush" of the checkpoint object(s) anyway... and noting that it is extremely unlikely that a checkpoint will require any more than a single such LogSegment... I think it very safe to live with this limitation.

Now back to the reason for the starvation callback... Some may recall that we ran a test where a bunch of files were written to in quick succession. No Flush()'s were invoked on any of these FileInodes because they were coming via SMB and, despite the client doing a Close(), a decision was made some time ago to not trigger a Flush() at the time of the Close(). And, as the current inode API has no notion of Close() itself, there really was no indication that we could go ahead and close the openLogSegments for these no-longer-being-accessed files.

So what would happen with this logic was if the connection pool was 1024, and the time to write to the first 1024 files in the sequence was short, when we get the first Write() to the 1025'd file, it'd have to wait for the 1st FileInode's 10 second MaxFlushTime to expire. Hence, it was deemed that this would be a good time to force the Flush(). And, indeed, the "upcall" was made (though not then called "starvationCallback") to trigger the Flush(). As a full Flush() involves updating metadata for a FileInode (other than just for the aforementioned chunked PUT contexts), it was necessary to obtain the DLM lock to ensure no fs client was attempting simultaneous access to the FileInode. Finally, at the time, package dlm was "above" package inode... so that meant the "upcall" needed to go all the way up to package fs... where it could obtain the DLM before basically making it look like a client called Close() in the FileInode.

It should be noted that subsequently package dlm was moved BELOW package inode... so the upcall need not go all the way up. And, indeed, the first version of this PR did just that... moving the upcall to target a func in package inode where the DLM was still obtainable.

Then along came an excellent observation by @charmster that we need not do a "full" FileInode Flush() at the time of Close()'ing a FileInode's chunked PUT context. Indeed, as previously described, we already had the FileInode Mutex there in the inode data structure just to enable us to Close() chunked PUT contexts "in the background" WITHOUT needing a Flush() call.

This PR makes exactly that change.

I hope this clears up the need for the starvationCallback()... and the correction to prior implementations of the range of code holding the FileInode Mutex.

I personally must admit to getting lazy and making use of "defer" to ensure all exits from a func that said "fileInode.Lock()" would have a corresponding "fileInode.Unlock()" at func exit. As should be obvious now, this was much too gross a code interval in which to hold the FileInode Mutex.

edmc-ss:
> I also suspect that the code in this Pull Request still contains a deadlock. The fact that we haven't hit a deadlock in testing doesn't mean it doesn't exist, it just means we haven't hit it. The fact that I can't spell out exactly how the deadlock would occur also doesn't mean it doesn't exist.

That's kind of a hard comment to respond to... :-)

That being said, I really like the specific example @charmster cited as a potential problem. But lets look at that sequence carefully. Indeed, there is logic in the "flusher" logic which deals with precisely this scenario. As you get into the "flusher", it checks if the openLogSegment field still references the chunked PUT context that it is being asked to Close(). Indeed, it is precisely the FileInode's openLogSegment field that indicates whether it still needs a Close()... so the test is trivial... And the only logic needed to prevent two contexts from doing the same test and both thinking "yes, I need to Close() it" is that very FileInode Mutex around the "test and set" of this openLogSegment field. Both starvationCallback() and the Flush() path ultimately call this very same "flusher".

I hope this verbally addresses at least this specific fear of a deadlock in this PR's current state.

But beyond that, to attempt to get back to the original charge "The fact that we haven't hit a deadlock in testing doesn't mean it doesn't exist,"... I'm currently in the process of amending this PR (possibly staged in a subsequent PR) to dramatically increase the automated "go test" func's in both fs and inode to hopefully give us much more confidence that deadlock risk has been addressed.

swiftclient/api.go
```diff
@@ -109,8 +127,8 @@ func ObjectDelete(accountName string, containerName string, objectName string, o
...
 
 // ObjectFetchChunkedPutContext provisions a context to use for an HTTP PUT using "chunked" Transfer-Encoding on the named Swift Object.
-func ObjectFetchChunkedPutContext(accountName string, containerName string, objectName string) (chunkedPutContext ChunkedPutContext, err error) {
-	return objectFetchChunkedPutContextWithRetry(accountName, containerName, objectName)
+func ObjectFetchChunkedPutContext(accountName string, containerName string, objectName string, useReserveForVolumeName ...string) (chunkedPutContext ChunkedPutContext, err error) {
```
edmc-ss: I've eliminated the need for the variadic... now requiring a single string useReserveForVolumeName argument... set to "" if not wanting to utilize the lone reserved conn.

charmster:
Let's see, there's a lot of comments to address ... i think i'll write a comment per comment.  from the Aug 13, 2018 10:17 pm comment:

the gyrations between chunked put connection context and the file `fileInode.Lock()` in this PR just strike me as bass-ackwards.  while the "fileInode.Lock()" lock is not held during the call to `swiftclient.ObjectFetchChunkedPutContext()`, it is held during the call to `swiftclient.SendChunk()` (which is fine), but its acquired during the starvation callback to close a chunked put connection.  i don't see an obvious bug there, but the layering feels unnatural and, therefore, suspect.

as i noted elsewhere, this PR puts chunked put connections above inode.Lock() in the locking heirarchy, wheras the "alternate PR" does not. (both approaches put chunked put connections above the inode's global.Lock() in the locking heirarchy, but that is less worrisome to me because we don't manipulate a chunked put context while holding `globals.Lock()`).

both approaches use `fileInode.Lock()` to protect the inode's `inFlightLogSegmentMap`, and both avoid holding the lock when doing a "background" flush of a full log segment.  (though the real key to that fix is to call `swiftclient.Close()` in a go routine rather then inline.)

this comment also criticizes my "alternate PR" because "the shared map approach requires a much more global Mutex... not just one confined to the particular FileInode in question imho. So I much prefer this PR's approach for that reason alone."

this is a bit misguided since this PR _introduced_ the acquisition of a global lock (globals.Lock()` in package inode) for each call to `doSendChunk()`, by the call to `openLogSegmentLRUTouch()` in the common case and `openLogSegmentLRUInsert()` in the uncommon case.

the changes I suggested in the "alternate PR" use the same lock for pretty much the same purpose, but require getting and releasing the lock _twice_ per call to `doSendChunk()` rather than _once_ per call in this PR.  i doubt that's a measurable performance issue, but if it were we could certainly change the inode code to avoid _any_ acquisition of the global lock in the common case (for either approach).

if it really is an issue we should also look at the usage of `globals.Lock()` in package fs, which is acquired for every `Write()` operation on the file by the call to `trackInFlightFileInodeData(), where it is acquired once for every call to `doSendChunk()`.  i imagine that optimizations useful in package inode would also work there.

charmster:
from the comment of Aug 13th at 10:40 pm ...

before this set of changes there was a real deadlock between the "chunked put context lock" (holding a chunked put context) and the fileInode.Lock(), since we held the fileInode.Lock() while trying to get the "chunked put context" lock and we tried to acquire the fileInode.Lock() if we ran out of chunked put context's (if an attempt to lock the "chunked put contenxt lock" would fail).  the fact that the fileInode.Lock() we tried to acquire was not the same inode as the one calling `swiftclient.ObjectFetchChunkedPutContextCnt()` didn't mean this could not be a deadlock.

but that deadlock has been fixed by this PR (and the "alternate PR").

there's a second deadlock, which @edmc-ss  describes in the comment referenced above.  up in the fs layer the code tracks "inflight log segments" and flushes them after 10 sec (unless something happens to trigger a flush earlier, like a call to `fs.Flush()`.  the flush occurs in `fs.inFlightFileInodeDataFlusher()` (or `fs.(*volumeStruct)inFlightFileInodeDataFlusher()` if you want to be pedantic).  that code gets the DLM lock for the inode, then calls down into the inode layer to `Close()` the current chunked put context associated with the file and wait for that and any others to finish flushing.

therefore release of the chunked put context via the in-flight file flusher mechanism first requires the DLM lock be acquired before the log segment can be released.  the DLM lock for that inode could be held by some other go routine which might be blocked trying to acquire just about any other lock in ProxyFS -- _because the DLM on an inode is, and remains, at the top of the locking heirarchy_.  in particular, that goroutine may try to acquire the headhunter lock (volume lock) to _fetch or update some metdata to headhunter_, for example in `fs.flushInodes()`.

if a checkpoint is in progress then the headhunter lock will be held by whatever goroutine is performing the checkpoint and that go routine, in turn, will nedd to acquire a chunked put context.  if all of the chunked put context's are in use then it will block waiting for one to become available.  if _all_ of the chunked put contexts are associated with inflight log segments and _all_ of those inodes have their DLM locks held by and are waiting for `fs.flushInodes()`, then we are truly deadlocked.

the strange thing is, in the sm-p1 case with 1024 connections in the pool, it would require 1024 inodes to have "open log segments" (which could happen) and all 1024 of those inodes would need to have operations in progress that hold the DLM lock while attempting to flush metadata to headhunter.  if just one of them was idle, the 10 sec should fire, flush the inode, release the chunked put segment, and progress would continue.  its hard to believe that happened with the kind of test workloads we're using!

regardless, the deadlock described just above is real.  fortunately, the other changes made in this PR (or the changes in the "alternate PR") fix the problem by allowing chunked put contexts to be `Close()` without having to get the DLM lock.

therefore, _there is no need for the changes in this PR to "reserve" one chunked put connection for each volume to avoid deadlock because this PR already fixes that deadlock_.

charmster:
and the comment of Aug 13, 2018 10:47 pm ...

@edmc-ss comments reference a description i wrote of a deadlock in the code that would exist if the code did not have a "reserved" chunked connection for headhunter flushes.  that description is wrong because it assumes that the `fileInode.Lock()` is held when getting the headhunter lock, which would occur if, say, the `fileInode.Lock()` were held when `inode.flushInodes()` is called *and* we didn't have a reserved chunked put connection for headhunter.  but fileInode.Lock() is not held in that case.

the DLM lock for an inode *is* held in analogous situation described in my comment of Aug 15, 2018 3:18 pm, and that would be a deadlock if it were not for the starvation callback mechanism in this PR (and the "alternate PR") *or* if it were not for the "reserved chunk connection" in this PR.

one problem with writing _clever code_ is that if it takes, say, 2 hours to write some code with fancy synchronization or locking, it will take me 4, 8, or 40 hours to really analyze to know whether there's a bug that causes deadlock or not.  its just hard.  and its also not fun.

over the years i've developed a coding philosophy when it comes to synchronization of having a small number of simple rules and following them scrupulously.  for example: define a locking hierarchy and being sure to follow it everywhere, without exceptions. using a lock to protect shared data structures and being sure to get it every time (even if there's some clever reason why it shouldn't be necessary).  etc.

i've also developed a certain intuition about code that may contain races, deadlocks, or data structure corruption.  my intuition is generally (but not always) correct.

some of the clever synchronization makes me nervous.  i'm not saying its wrong, but i'm not confident that its right.

PR #175, opened 2018-08-21
Add ramswift shutdown to fs & inode testTeardown() func's

edmc-ss:


bschatz-swift (commented):
Looks good to me - I am testing it now

PR #179, opened 2018-08-23
Etcd prototype files

bschatz-swift:
Read the README for tips


etcd/Vagrantfile
```diff
@@ -0,0 +1,24 @@
...
+N = 3
+Vagrant::configure("2") do |config|
+    (0..N-1).each do |machine_id|
+        config.vm.box = "centos-74-minimal-20171228"
+        config.vm.box_url = "https://o.swiftstack.org/v1/AUTH_misc/vagrant_boxes/centos-74-minimal-20171228.box"
+        config.vm.synced_folder ".", "/vagrant", type: "virtualbox"
+        config.vm.define "etcd#{machine_id}" do |machine|
+            machine.vm.hostname = "etcd#{machine_id}"
+            machine.vm.network "private_network", ip: "192.168.60.#{10+machine_id}", :name => "vboxnet0", :adapter => 2
+            if machine_id == N-1
```
edmc-ss, line 14: Not sure why this is only to be done on the "last" machine_id...

bschatz-swift: I grabbed a lot of this file off the web and thought it was throw away code for now so I did not look at it in detail.

I went back and looked at it and I could not understand it completely.   I want the vagrant_bootstrap.sh script to run on all nodes to make sure that we have all updates.  I have since moved that outside the check.

We only want the ansible section to run after all nodes are configured since it verifies that etcd is up and running.   We do not want that to happen every loop

bschatz-swift:
This is very old....

PR #180, opened 2018-08-28
Scheduled snapshots

edmc-ss:


edmc-ss:
Ok... so with the latest `git push`, this PR is now complete and ready for final review and integration. Please comment as you see fit.

If you want to "try it out", you can go into proxyfsd/file_server.conf and uncomment (presumably) the second of the two SnapShotPolicy: lines under [Volume:CommonVolume]. That will turn on the SnapShot Policy mechanism.

You can then go to `/.snapshot/` and see your SnapShot's created and pruned per the policy. Of course the finest granularity of taking scheduled SnapShot's being a minute, watching the pot of water boil is a bit tedious.

Alternatively, you can see this same list of SnapShot's via the embedded HTTP Server inside ProxyFS. There, you can even create (and delete) SnapShot's manually as well. Using this mechanism, you could presumably create a sequence of SnapShot's more rapidly than once a minute... just follow the timestamp naming convention (RFC3339) to "impersonate" your chosen SnapShotPolicy/Schedule... and watch the "prune" function happen. Do note, however, that "pruning" only occurs at the time the SnapShotPolicy daemon() wakes up to perform a scheduled SnapShot.

edmc-ss:
Hmm... very curious issue...

The timezone my Mac is in is America/Los_Angeles... same as the two example SnapShotPolicy's I included in proxyfsd/file_server.conf... so everything worked as expected.

But then when running on my SAIO VM, where the timezone is set to UTC (i.e. different than either SnapShotPolicy's TimeZone: value), no snapshots get scheduled. This is no doubt because I am using the time.Equals() func that explicitly claims as `not equal` two times that differ only in UTC (e.g. 2pm UTC and 3pm UTC+1 are reported as `not equal`).

I need to fix this since we cannot be sure the nodes on which proxyfsd is running will be in the same timezone as the SnapShotPolicy's...

edmc-ss:
OK, this is getting really strange. Here is a `date` command sequence issued in relatively quick succession inside the SAIO VM (presumably has no local time zone setting so uses UTC btw):
```
[vagrant@localhost ProxyFS]$ date
Thu Aug 30 19:38:35 UTC 2018
[vagrant@localhost ProxyFS]$ date
Thu Aug 30 19:49:34 UTC 2018
[vagrant@localhost ProxyFS]$ date
Thu Aug 30 20:21:06 UTC 2018
[vagrant@localhost ProxyFS]$ date
Thu Aug 30 20:21:25 UTC 2018
```
Yes, that is a 31+ minute jump between the 2nd and 3rd.

So there are two issues (one real, one "virtual") to yet solve:
1) Adjust time comparisons in the SnapShotPolicy daemon() to use the specified (or defaulted) TimeZone, and
2) Ensure that arbitrary jumps in time due to poor time/clock emulation in e.g. VirtualBox (qemu) doesn't "upset" SnapShotPolicy daemon().

edmc-ss:
Ok - latest commit resolves (1)... and (2) turns out to resolve itself with time.

PR #182, opened 2018-08-31
Apply SIGHUP Pause/Resume-enforcing "gate" to package fuse

edmc-ss:
Much like jrpcfs, package fuse is the top-most entrypoint for
proxyfsd and, as such, bears the responsibility to ensure it
follows the convention that, once paused, no traffic to the
layer below it (package fs in this case) will receive any
requests... and that this protection continues until having
completed the resume phase.

### DIFF
--- a/fuse/config.go
+++ b/fuse/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path"
+	"sync"
 	"syscall"
 	"time"
 
@@ -30,6 +31,9 @@ type mountPointStruct struct {
 }
 
 type globalsStruct struct {
+	gate sync.RWMutex // SIGHUP triggered confMap change control
+	//                   API Requests RLock()/RUnlock
+	//                   SIGHUP confMap changes Lock()/Unlock()
 	whoAmI        string
 	mountPointMap map[string]*mountPointStruct // key == mountPointStruct.mountPointName
 }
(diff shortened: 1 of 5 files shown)


edmc-ss:
I've created this PR branched off of 1.7.1 as a proposed hotfix to resolve the NFS-only rsync CHO failures. Once we nail down reproduction steps on stock 1.7.1, I'd like to try those steps against this PR to verify the fix.

fuse/config.go
```diff
@@ -134,6 +138,8 @@ func PauseAndContract(confMap conf.ConfMap) (err error) {
 		return
 	}
 
+	globals.gate.Lock()
+
```
charmster, line 142: there's a batch of "error out" cases below where we call return without dropping the lock. 

i'm assuming if we return an error we won't call ExpandAndResume() so perhaps its a moot point as to whether we release it now ... if we don't call ExpandAndResume() then the only other course of action is to exit ...

Hmmm.  the error cases in ExpandAndResume() probably have the same issue.  If we exit before we release the gate, who will release it?

Or does this not matter at all?

bschatz-swift (approved):

charmster (approved):
Looks good.

There is an issue with releasing the "gate" in error cases in PauseContract()/ExpandResume() but its not really important i think.

fuse/config.go
bschatz-swift: Good point!   Probably it should be a defer Unlock()?

edmc-ss:
Leaving this branch alive so as to easily find the commit that is a candidate for a 1.7.1 "hotfix".

PR #210, opened 2018-10-11
Added new transitions package (unintegrated)

edmc-ss:
This new package provides a API and a callback mechanism designed to
entirely replace the repetitive code sequence in each main() func
and most of the _test.go files that have to startup (and shutdown)
the various packages they are dependent upon... and do so in the
proper order. Maintenance of these dependencies and sequences will
eventually be totally automated once each such package implements
an init() func to register their callbacks and the main() (and
test) func's replace their package initialization/shutdown sequences
with a simple set of calls to the transitions package.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ gopkgsubdirs = \
 	stats \
 	statslogger \
 	swiftclient \
+	transitions \
 	utils \
 	version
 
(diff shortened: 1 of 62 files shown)


edmc-ss:
Note that, while this now builds (and quicker, too), it's not yet ready to merge.
I'm actually only done up through package inode... still 4 more to go... along
with the various programs that call into them. So far - so good!

edmc-ss:
As mentioned in the now single/collapsed commit, package transitions integration is now complete except for the as-yet-unused pfsalived program. As such, I feel ready to merge this to development at any point.

Note that a parallel PR on the proxyfs-jrpc-client repo must also be merged to it's development branch simultaneously.

edmc-ss:
And with the last rebased/collapsed commit, package transitions is now fully integrated !!!

PR #214, opened 2018-10-27
Fix python lint errors.

charmster:
tox lint is suddenly complaining about binary operators at the end
of the line.  Change pfs_middleware/tox.ini to ignore that warning.

A regular expression in pfs_middleware/pfs_middleware/utils.py used
double quotes for the RE without escaping `\d`.  Change it to use
a raw string.


charmster:
This does seem to fix the problem!

Issue #188, opened 2018-09-10
Add support for end_marker query parameter in container listings


tipabu:

tipabu:
Closed via https://github.com/swiftstack/ProxyFS/pull/227.

PR #227, opened 2018-12-05
pfs_middleware: Support `end_marker` query params

### LINKED ISSUE #187: Add support for end_marker query parameter in account listings

tipabu:
... for both accounts and containers. Fixes #187, #188.

Upgrade considerations are minimal: old pfs_middleware talking to new JRPC server won't include the new arg, so it'll be populated with an empty string. New pfs_middleware talking to old JRPC server includes the arg but the server ignores it, which is no worse than the old behavior.


edmc-ss (approved):
Looks great!

PR #229, opened 2018-12-07
Add an escape hatch for account owners

tipabu:
Let swift_owner users look at raw ProxyFS data by including an

    X-Bypass-ProxyFS: true

header. **Don't** let them modify anything – it would almost certainly lead to filesystem corruption.

This may be useful to allow read-only proxyfs instances to run outside of the rest of the cluster.

Maybe this should _also_ require a filter config option to enable? Something to think about.


pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -709,8 +709,18 @@ def __call__(self, req):
                     if denial_response:
                         return denial_response
 
-                # Authorization succeeded; dispatch to a helper method
+                # Authorization succeeded
                 method = req.method
+
+                # Check whether we ought to bypass. Note that swift_owner
+                # won't be set until we call authorize
```
tipabu, line 726: Hurray for functional testing! Previous version tried to do this immediately once we get into `__call__`, which flat-out didn't work.

There's still a complication, however: trying to GET an object seems to use the default storage policy, rather than the appropriate policy for the container -- pretty sure this is because `get_container_info` calls through `make_pre_authed_env` which drops in a no-op authorize callback, so we don't know that the user is a swift_owner. Eventually, this leads to us caching the container's non-existence (!) and we just assume the default storage policy...

tipabu:
I *think* I'm happy with this now? Clearly benefits from someone actually trying to exercise it. Not sure how long it would take for @edmc-ss to hack up a proxyfsd that could work with a remote endpoint though :wink:

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -709,8 +709,25 @@ def __call__(self, req):
...
+                if config_true_value(req.headers.get('X-Bypass-ProxyFS')) and \
+                        req.environ.get('swift_owner'):
+                    if method not in ('GET', 'HEAD'):
```
edmc-ss, line 729: While the point about how trivial it is to muck up the underlying file system metadata, I nevertheless think we should allow PUT here as well. Perhaps we could restrict it to not allow PUTs in the Checkpoint Container... but I'm not sure pfs_middleware knows what that is. Perhaps it should...

edmc-ss (changes requested):
As commented, I'd like to enable PUTing of LogSegments too.

So long as such PUTs never overwrite existing Objects... and never get overwritten by normal ProxyFS activity, there is no potential for corruption. Limiting PUTs to LogSegment Containers would be a "nice to have" but so long as clients are being trusted anyway, that's probably not necessary.

edmc-ss (approved):
Upgrading to Approved... we can iterate on this later.

PR #230, opened 2018-12-07
Add ability for account owners to get read plans for objects

tipabu:
... by adding a `?get-read-plan` query parameter to GET requests.

Note that the read plan respects any `Range` header that may have been included in the request.

This kinda goes hand-in-hand with #229. If the escape hatch is further guarded by a config option, this should be, too. Still not entirely sure about the API.


pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -1522,6 +1522,9 @@ def get_object(self, ctx):
 
         read_plan, raw_metadata, size, mtime_ns, ino, num_writes, lease_id = \
             rpc.parse_get_object_response(object_response)
+        if req.environ.get('swift_owner') and 'get-read-plan' in req.params:
+            return swob.HTTPOk(request=req, body=json.dumps(read_plan),
+                               headers={"Content-Type": "application/json"})
```
tipabu, line 1563: Maybe I should use something like `application/proxyfs-read-plan+json`? Or include a header like `X-ProxyFS-Read-Plan: True`? I kinda want _something_ to differentiate from when you're not actually looking at a bimodal account...

edmc-ss:
This looks great... though is there a more conventional HTTP mechanism? I'm not an HTTP person but the Accept header comes to mind. Or does that just refer to how the data is returned (e.g. JSON versus text versus...)? In the GetReadPlan case, the data isn't actually being returned of course.

I believe we should also talk about "leases" now... it would be somewhat helpful to indicate to the client how long a returned ReadPlan may be cached. In the near term, we wouldn't know what to set it to and, indeed, it's even possible it's already stale by the time pfs_middleware returns it.

pfs_middleware/pfs_middleware/middleware.py
edmc-ss: Ah, yes, here you've already responded to the comment I just made :-).

edmc-ss (approved):
Looks great to me other than deferring to the Accept vs. query-parm vs. ??? decision

tipabu:
> This looks great... though is there a more conventional HTTP mechanism? I'm not an HTTP person but the Accept header comes to mind. Or does that just refer to how the data is returned (e.g. JSON versus text versus...)?

Yeah, `Accept` is about content negotiation and figuring out what type the response body ought to have. I see no reason to mess with that here; read plans will just **be** JSON, in much the way that SLO manifests just **are** JSON.

> I believe we should also talk about "leases" now... it would be somewhat helpful to indicate to the client how long a returned ReadPlan may be cached. In the near term, we wouldn't know what to set it to and, indeed, it's even possible it's already stale by the time pfs_middleware returns it.

IDK... seems like the client ought to be in charge of its own caching strategy... or are we thinking about the server _responding from cache_, and exposing _that_ cache information? Maybe there'd be some value there...

PR #231, opened 2018-12-11
Add support for X-Object-Sysmeta-Container-Update-Override-Etag

tipabu:
... so objects (such as SLOs) can override their container listing entries.

This should fix up some failing bimodal tests.


tipabu:
@gerardgine says it fixes things, so I'm'a merge it 😉

PR #240, opened 2019-01-07
pfs_middleware: get-read-plan follow-ups

tipabu:
* Include object metadata in read plan response
* Include `X-Proxyfs-Read-Plan` header as confirmation that we are talking to a ProxyFS account and not generic Swift
* Check value of get-read-plan query param; allow it to be explicitly false


notmyname (approved):
Looking at the code (but not running it), it looks good. I like the implementation of the `get-read-plan` query parameter, the tests look to cover the variations, and I think the response header indicating the read plan is good.

charmster (approved):
It looks good (though i really don't understand middleware well enough to say that).  i like that you test the range operation, though i don't think "sacrocyst" is a word (but "sacrocystis" is).

PR #241, opened 2019-01-07
pfs_middleware: Add config option for bypass/get-read-plan

tipabu:
We don't necessarily want to expose all of ProxyFS's internals to the world, even in some read-only mode. Add a `bypass_mode` config option that defaults to `off` (to reflect pre-bypass behavior) but may be set to `read-only` or `read-write`.

### DIFF
--- a/pfs_middleware/pfs_middleware/middleware.py
+++ b/pfs_middleware/pfs_middleware/middleware.py
@@ -647,6 +647,11 @@ class PfsMiddleware(object):
         # a few extra for JSON quotes, commas, et cetera.
         self.max_coalesce_request_size = self.max_coalesce * 1100
 
+        self.bypass_mode = conf.get('bypass_mode', 'off')
+        if self.bypass_mode not in ('off', 'read-only', 'read-write'):
+            raise ValueError('Expected bypass_mode to be one of off, '
+                             'read-only, or read-write')
+
     @swob.wsgify
     def __call__(self, req):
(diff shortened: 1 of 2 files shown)


pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -725,8 +730,10 @@ def __call__(self, req):
                 # Check whether we ought to bypass. Note that swift_owner
                 # won't be set until we call authorize
                 if config_true_value(req.headers.get('X-Bypass-ProxyFS')) and \
+                        self.bypass_mode in ('read-only', 'read-write') and \
                         req.environ.get('swift_owner'):
-                    if method not in ('GET', 'HEAD'):
+                    if self.bypass_mode == 'read-only' and method not in (
+                            'GET', 'HEAD'):
```
charmster, line 733: Hmm.  I think we should disallow DELETE even in the read-write case.  That way if a ProxyFS object turns up missing we don't have to worry that a "client" might have deleted it using this interface.

that does create an issue if a client creates an object but decides, for whatever reason, that it doesn't want to register the object with Proxyfs, it will not be able to delete it.

Can we do anything to prohibit the over-write of existing objects via this interface?  An object should never be updated once its created.

charmster (commented):
I have a concern about a client DELETE-ing objects or over-writing objects when bypass_mode == 'read-write'.  I'd like to see us prevent both, if possible.

(I'm not sure how valid my concern is.)

PR #261, opened 2019-02-14
Always provide JSON listings on new-enough Swift

tipabu:
On new enough swift, we should always send JSON listings; see https://github.com/openstack/swift/commit/4806434


tipabu:
On development as https://github.com/swiftstack/ProxyFS/commit/f5e2b2d1850c7ab6a24f4aab5cb69ea7cd77880b

PR #264, opened 2019-02-28
Re-implemented locking schemes for paths needing multiple locks

edmc-ss:
Primarily bi-modal requests (but also Rename) often require obtaining
multiple locks. As locking order may not always be easily deterministic,
avoiding deadlocks requires a "try-lock" approach. Many of these fs
APIs had unique implementations making it challenging to undertstand
interactions with other multi-locking APIs.

A key ingredient of several of these APIs was the fs.resolvePath() helper
func. This helper has now been significantly generalized to be useful
across the entire spectrum of fs APIs. Each of the "try-lock'ing" fs APIs
has also now been re-worked to leverage fs.resolvePath() as well.

Finally, a notable omission in fs.Rename() has been corrected. A code
review revealed that the target path of a Rename() that pre-exists
was not locked. Hence, if one has two files A and B... and Rename()'s
A to B, what happens is that the LinkCount on B's inode is decremented
without holding on to an Exclusive Lock at the time. Indeed, if the
LinkCount now drops to zero, code will be invoked to actually delete
the inode... all while a potentially 2nd thread of execution is
acting on it. The simplest example of conflict might be this 2nd thread
doing a Link() mapping C to B's inode... in the process incrementing
the LinkCount. These two LinkCount modifications are not atomic, hence
the resultant LinkCount could be correct, one too low, or one too high.
The new generalized fs.resolvePath() is now used to obtain all locks
during fs.Rename() to avoid this race condition.

A notable impact of this change involves the fs.MiddlewareGetContainer()
logic. Given the need to "flatten" the directory tree underneath a
Container (i.e. underneath a subdirectory of '/'), the necesssary
treewalk must descend into encountered subdirectories. The point at
which it does this is just after recording the existence of each
subdirectory. Hence, a subdirectory (of the selected Container) named
foo that contains a file named bar will be listed as:

    foo
    foo/bar

Now suppose there is a file named foo.suffix in the same Container
adjacent to the subdirectory foo. In this case, the treewalk
algorithm will encounter foo.suffix AFTER it has already descended
into subdirectory foo. Hence, the Container listing will be:

    foo
    foo/bar
    foo.suffix

Note that '.' should sort BEFORE '/'. Indeed a non-bimodal Swift
Container having three Objects named "foo", "foo/bar", and "foo.suffix"
would instead be listed:

    foo
    foo.suffix
    foo/bar

While it is somewhat feasible to avoid this deviation from sort order,
it is very problematic when considering the impact to the implementation
supporting prefix, marker, and endmarker (options available for a GET
on a Swift Container).

This change does not attempt to avoid this anomaly.


charmster:
copying in a comment i posted on Slack ...

the question of "/" sort order is interesting.  there are only  a small number of characters that sort before "/" in ASCII, though "." and "," (comma) are the most problematic.  aside from that, there are 32 control characters (^A, ^B, ... etc.), the 15 printable characters are `[ !#$%&' ()*+,-.]` (space is the first character, also bang, sharp, dollar, percent, and, single quote, left and right parent, splat, plus, comma, dash, period).

as long as we are consistent about the sort order, then markers used for directory (container) traversal should work OK.  will this bug any clients?

i'm not sure there's a problem with prefix, other than the order of entries.  if the client says `prefix="bar/foo"` then they it gets everything in the container that starts with "foo" (with the sort order issue intact).  if the client says `prefix="bar/foo/"` then it gets everything under the directory "foo", but not "bar.suffix".

where does Unicode fit into this question?  are we sorting by byte values or Unicode code points?

curiously (actually, i'm sure it was by design), if two byte arrays represent two different unicode characters, a byte by byte comparison will always sort the arrays by code point order.

fs/api_internal.go
```diff
@@ -1515,319 +1515,757 @@ Restart:
...
+	// Setup shortcuts/contants
+
+	dlmCallerID = dlm.GenerateCallerID()
+	inodeVolumeHandle = mS.volStruct.inodeVolumeHandle
+
+	// Compute initial response
+
+	restartBackoff = time.Duration(0)
+
+Restart:
```
charmster, line 1916: shouldn't labels be lower case?

edmc-ss: Interesting... didn't know there was a convention.
Indeed, I was just trying to follow the pattern of other label gotos both in this file and elsewhere.
Happy to adopt your preferred naming convention.
There are probably only a handful of other files that actually use goto's :-)

edmc-ss:
Awaiting a cleaner run thru Ci Swift/Swift3 functional tests before merging this in...

PR #266, opened 2019-03-04
Enabled 64-bit atomic operations on 32-bit architectures

edmc-ss:
All atomic.*64() APIs simply won't work for arm7l
    Package bucketstats is the only user of these
    As such, platform-specific Uint64-accessing calls
    to sync/atomic func's are introduced that are
    equivalent to what was called for 64-bit builds
    but wrap a sync.Mutex lock around the equivalent
    operations on 32-bit architecture builds.

The mechanism to obtain memory size under Linux is
    consistent between 32-bit and 64-bit architectures
    but the struct element definition is not. As such,
    a 32-bit version that "up-casts" to a uint64 was
    introduced.

Various other operations were directly made supportive
    of both 32-bit and 64-bit builds.


edmc-ss:
Please note that I also tweaked some stress tests in fs to ensure they work just as hard but don't consume excessive amounts of memory. Those tests now can run inside a 2GB VM/Container.

PR #269, opened 2019-03-12
Revived missing SnapShotPolicy.daemon() in package inode

edmc-ss:
Somehow, this was missing from the ServeVolume()/UnserveVolume() callbacks.
As such, no "scheduled" SnapShots were being taken.

Not sure how/when the lauching was lost (probably in the move to the new
package transitions model)... but this commit resolves that regardless.


edmc-ss:
Please disregard the typo in this short-lived branch's name :-)

PR #270, opened 2019-03-13
Changed SnapShotPolicy to respond to SIGHUP

edmc-ss:
Previously, a change to a Volume's SnapShotPolicy specification would
only be picked up on a Volume newly served (including restart of course).
This change makes to interpretation of the SnapShotPolicy occur on SIGHUP
as well.

In addition, this change also halts the taking of SnapShots based on an
existing SnapShotPolicy during SIGHUP handling (that also occurs during
the shutdown sequence).

### DIFF
--- a/inode/config.go
+++ b/inode/config.go
@@ -596,12 +596,6 @@ func (dummy *globalsStruct) ServeVolume(confMap conf.ConfMap, volumeName string)
 		return
 	}
 
-	err = volume.loadSnapShotPolicy(confMap)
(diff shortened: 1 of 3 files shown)


edmc-ss:
Note that I also updated the 1.10.0 Release Notes (posthumously) to mention that the SnapShotPolicy feature had been inadvertently removed in 1.9.5 and was reinstated in 1.10.0 (although not ideally as this PR updates).

PR #275, opened 2019-03-27
pfs_middleware: test under py3

tipabu:
There's also some reorganization of tox environments, just to help cut down on definitions. Add tox envs for py36 and py37, testing against both the "current release" and Swift master. Note: our minimum-supported Swift version (2.9.0) can't even be imported on py3.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ pre-generate:
 
 python-test:
 	cd meta_middleware && tox -e lint
-	cd pfs_middleware && tox -e py27,py27-old-swift,lint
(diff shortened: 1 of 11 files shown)


tipabu:
All right, I *think* that'll work. Note that I had issues running `run_docker_tests.sh` locally; after a bunch of `go test -vet all ...` lines, it'd bomb out with

    2019/03/27 16:12:19 mount helper error: fusermount: fuse device not found, try 'modprobe fuse' first
    2019/03/27 16:12:19 mount helper error: fusermount: fuse device not found, try 'modprobe fuse' first

Rewriting `launch.sh` to run `make python-test` instead of just `make` had things passing, though.

PR #276, opened 2019-03-29
pfs_middleware: Add PROXYFS method

### LINKED ISSUE #245: Expose JSON-RPC interface to proxyfsd via pfs_middleware
This is part of the general move to expose more of ProxyFS directly to clients; see also: #229, #230, #241

Introduce a new HTTP method (`PROXYFS`?) that can be used on the volume/account to allow clients to make JsonRpcClient requests directly. Method will only be valid at the account (and only for bimodal accounts), so we can use that to look up which proxyfsd instance to talk to. Require JSON request bodies, probably with some limit on size – send it straight to proxyfsd. Emit the JSON response verbatim back to the client. If the call times out, respond `502 Bad Gateway`.

Similar to the `X-Bypass-ProxyFS` header, this should be limited to Swift owners; once we have a `bypass_mode` option, it should (as a first pass) require `read-write`.

Bonus points:
- parse requests to verify format before sending to proxyfs
- whitelist of RPC methods to allow when `bypass_mode = read-only`
- allow client to specify an RPC timeout (keeping some reasonable default if _not_ specified)
- consider allowing multiple JSON objects per request, instead of requiring pipelined HTTP requests
- consider sniffing for errors and responding 400

tipabu:
Only accessible to swift owners. Disable with `bypass_mode = off`.

Lets REST clients make arbitrary RPCs against proxyfsd. Provides some degree of validation before making said request, including separate whitelisted RPC methods for `read-only` vs. `read-write`.

Closes #245.


pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -855,6 +863,90 @@ def _fetch_appropriate_acl(self, ctx):
...
+            except (TypeError, KeyError, ValueError):
+                self.logger.debug('Error parsing payload #%d %r',
+                                  i, json_payload, exc_info=True)
+                return swob.HTTPBadRequest(
+                    request=req,
+                    body=('Could not parse/validate JSON payload %r' %
+                          (json_payload, )))
+            payloads.append(payload)
+
+        # TODO: consider allowing more than one payload per request
```
edmc-ss, line 916: Indeed, for performance in some heavily traffic'd pfsagentd deployments, I imagine multiple requests being issued via FUSE to it... if we get multiple around the same time, and for the same Account, we could coalesce them in a single HTTP PROXYFS request.

That being said, it's not like FUSE makes that easy to determine. If we have some sort of "throttling" on the pfsagentd side (e.g. only send 'N' at a time), I could imagine a work queue of the requests that are >N that we've sent... meaning we could send some or all of them in one HTTP PROXYFS Request.

Not an optimization I'm looking to do unless/until it's performance advantage is helped.

On that note, I'd frankly rather see Swift adopt HTTP/2... and ultimately HTTP/3 (QUIC) to achieve such improvements rather than just for the HTTP PROXYFS method, no?

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -855,6 +863,90 @@ def _fetch_appropriate_acl(self, ctx):
...
+                "Timeout (%.6fs) communicating with %s, calling %s",
+                self.proxyfsd_rpc_timeout, ctx.proxyfsd_addrinfo,
+                payloads[0]['method'])
+            return swob.HTTPBadGateway(request=req)
+        except socket.error as err:
+            self.logger.debug("Error communicating with %r: %s.",
+                              ctx.proxyfsd_addrinfo, err)
+            return swob.HTTPBadGateway(request=req)
+        else:
+            if 'error' in response:
```
edmc-ss: I don't think I've got a simple way to *NOt* include the error field in a response.
I realize the JSON RPC 2.0 spec says something like "only error or payload" can be included. The Go JSON Marshaler doesn't make that very easy. It's possible... but we never did that in the package jrpcfs layer. I *could* go re-write all of that... but don't want to do that now.

Indeed, I'd much sooner re-write it by ditching JSON RPC altogether. It's crap imho.

edmc-ss: I guess what I am saying is that an "empty" error in the response should also be allowed.

Heck, for all I know, the JSON RPC layer we are leveraging takes care of this though. Do you know?

pfs_middleware/pfs_middleware/utils.py
```diff
@@ -133,7 +135,7 @@ def call(self, rpc_request, timeout):
                             % (err, line))
 
                 errstr = response.get("error")
-                if errstr:
+                if errstr and raise_on_rpc_error:
```
edmc-ss, line 138: Can we "adorn" this error/exception report with, say, a "PROXYFS" tag so that it is easy to search for in the log? As written, each individual error would only share "Error in" in common... I dare say there may be other lines with those two words in them, no?

edmc-ss (commented):
More generally, I didn't know if this PR already incorporates the "X-Bypass-Proxyfs" (for LogSegment GETs and PUTs). Is there another PR I need to grab (first) to get that... or is it in development already?

pfs_middleware/pfs_middleware/middleware.py
tipabu: If/when we want to do this, I think a lot of the groundwork has already been laid -- should be a matter of adding a loop around L923 and having it yield out bodies.

HTTP/2 is interesting for sure, but I'm not clear on how it would map to the WSGI model -- I guess maybe we just stand up an nginx instance or something to talk HTTP/2 to the client while spinning up a bunch of HTTP/1 connections to Swift? _\*shrug\*_

pfs_middleware/pfs_middleware/middleware.py
tipabu: I *think* it might be OK? I thought I cribbed this pattern from somewhere... but no, it looks like we've got

            errstr = result.get("error")
            if errstr:
                errno = utils.extract_errno(errstr)
                raise utils.RpcError(errno, errstr)

down in `_rpc_call`. I'm fine with changing it to `if response.get('error'):`. #willfix

pfs_middleware/pfs_middleware/utils.py
tipabu: Are you thinking of just the new `PROXYFS`-method requests, or all `RpcError`s generally?

The point of adding the `raise_on_rpc_error` arg is that `PROXYFS` requests won't raise these errors at all -- they'll just return 422 with a JSON body that (should) exactly match the response from proxyfsd.

When the "normal" methods fail in such a way as to trip an `RpcError`, that should get caught and logged around https://github.com/swiftstack/ProxyFS/blob/1.10.0/pfs_middleware/pfs_middleware/middleware.py#L787-L802 where it'll pick up `RPC error:`, which should be rather unique to ProxyFS...

tipabu (commented):
> More generally, I didn't know if this PR already incorporates the "X-Bypass-Proxyfs" (for LogSegment GETs and PUTs). Is there another PR I need to grab (first) to get that... or is it in development already?

Yeah, the `X-Bypass-Proxyfs` should already be included. That came in on development as https://github.com/swiftstack/ProxyFS/commit/89e9cd2 (for read-only) and https://github.com/swiftstack/ProxyFS/commit/050b105 (for read-write).

pfs_middleware/pfs_middleware/middleware.py
edmc-ss: Sounds fun at least... certainly as interesting as Py2->Py3 work, no? :-)

pfs_middleware/pfs_middleware/middleware.py
edmc-ss: I'm sure you're right... in thinking about it, that funky JSON RPC layer we are using in ProxyFS does all sorts of stuff for us. It codes the event loop, serializes/deserializes, etc... all we do is implement a bunch of functions that, for e.g. Server.RpcFooRequest/Reply, we make a public function named Foo() that takes two by-reference args: in *FooRequest struct, and reply *FooReply struct. That's it... so this must be ok...

pfs_middleware/pfs_middleware/utils.py
edmc-ss: I *was* thinking of just the new PROXYFS method... but take your point.

Let me work for it as is... and see how it goes.

edmc-ss:
I'll create a branch of my branch to see how this works so far...

edmc-ss:
Thanks Tim - this is perfect!

I've cherry-picked your squashed commit (already had the rpc.py fix) onto my pfsagentd branch... so you can delete this PR at your preference.

Marching forward!

PR #277, opened 2019-04-03
Added provisioning of Python 3.6 for runway & SAIO

edmc-ss:


tipabu (approved):
Seems to make sense.

So how many different environments do we have to provision for in this repo? We've got the Dockerfile in test/container, these two... anywhere else? I know the Dockerfile is used (indirectly?) by Travis for the PR checks -- who/what/where uses the cookbook vs the shell script, and why?

edmc-ss:
I dunno... Maybe:
- Travis CI (ProxyFS/.travis.yml)
- CI (@gerardgine knows what, if anything, additional)
- runway (ProxyFS/cookbooks...)
- SAIO (ProxyFS/saio)
- Native MacOS (although here only Go code is built/tested)
- Native Linux devs (perhaps @charmster could answer)
- SwiftDebBuilder (it actually disables most of the testing... just builds the .rpm)

PR #291, opened 2019-05-07
Bugfix pfs coalesce

charmster:
Fix bugs with the COALESCE operation for static-large objects where it dropped the metadata (headers) associated with the operation on the floor and failed to return the correct modification time.

Lack of headers broke multipart-upload in the S3API because the new object lacked the ACL headers so it could not be accessed via S3API.  It also dropped the S3API and SLO ETag values on the floor, which meant the object did not return the correct ETag later.

First commit: store selected metadata associated with the operation.  Mung the ETag headers os the request values are returned until the object is modified via SMB or NFS, at which point the ProxyFS generated ETag is returned.

Second commit: return the correct modification time for the COALESCE operaiton.

### DIFF
--- a/inode/file.go
+++ b/inode/file.go
@@ -915,6 +915,9 @@ func (vS *volumeStruct) SetSize(fileInodeNumber InodeNumber, size uint64) (err e
 		return
 	}
 
+	// changing the file's size is just like a write
+	fileInode.NumWrites++
+
 	err = fileInode.volume.flushInode(fileInode)
 	if nil != err {
 		logger.ErrorWithError(err)
@@ -1010,7 +1013,9 @@ func (vS *volumeStruct) resetFileInodeInMemory(fileInode *inMemoryInodeStruct) (
 	return
 }
 
-func (vS *volumeStruct) Coalesce(destInodeNumber InodeNumber, elements []*CoalesceElement) (coalesceTime time.Time, numWrites uint64, fileSize uint64, err error) {
+func (vS *volumeStruct) Coalesce(destInodeNumber InodeNumber, metaDataName string, metaData []byte, elements []*CoalesceElement) (
+	coalesceTime time.Time, numWrites uint64, fileSize uint64, err error) {
+
 	var (
 		alreadyInInodeMap                  bool
 		destInode                          *inMemoryInodeStruct
@@ -1131,7 +1136,7 @@ func (vS *volumeStruct) Coalesce(destInodeNumber InodeNumber, elements []*Coales
 
 	for _, element = range elements {
 		elementInode = inodeMap[element.ElementInodeNumber]
-		destInode.NumWrites += elementInode.NumWrites
+		destInode.NumWrites += 1
 		elementInodeExtentMap = elementInode.payload.(sortedmap.BPlusTree)
 		elementInodeExtentMapLen, err = elementInodeExtentMap.Len()
 		for elementInodeExtentMapIndex = 0; elementInodeExtentMapIndex < elementInodeExtentMapLen; elementInodeExtentMapIndex++ {
@@ -1187,15 +1192,18 @@ func (vS *volumeStruct) Coalesce(destInodeNumber InodeNumber, elements []*Coales
 	}
 
 	// Now, destInode is fully assembled... update its metadata & assemble remaining results
-
 	coalesceTime = time.Now()
-
 	destInode.CreationTime = coalesceTime
 	destInode.AttrChangeTime = coalesceTime
 	destInode.ModificationTime = coalesceTime
 
-	numWrites = destInode.NumWrites
+	// attach new middleware headers (why are we making a copy?)
+	inodeStreamBuf := make([]byte, len(metaData))
+	copy(inodeStreamBuf, metaData)
+	destInode.StreamMap[metaDataName] = inodeStreamBuf
 
+	// collect the NumberOfWrites value while locked (important for Etag)
+	numWrites = destInode.NumWrites
 	fileSize = destInode.Size
 
 	// Now, destInode is fully assembled... need to remove all elements references to currently shared LogSegments
--- a/inode/file.go
+++ b/inode/file.go
@@ -993,7 +993,6 @@ func (vS *volumeStruct) resetFileInodeInMemory(fileInode *inMemoryInodeStruct) (
 	)
 
 	fileInode.dirty = true
-
 	fileInodeExtentMap = fileInode.payload.(sortedmap.BPlusTree)
 
 	ok = true
@@ -1013,11 +1012,13 @@ func (vS *volumeStruct) resetFileInodeInMemory(fileInode *inMemoryInodeStruct) (
 	return
 }
 
-func (vS *volumeStruct) Coalesce(destInodeNumber InodeNumber, metaDataName string, metaData []byte, elements []*CoalesceElement) (
-	coalesceTime time.Time, numWrites uint64, fileSize uint64, err error) {
+func (vS *volumeStruct) Coalesce(destInodeNumber InodeNumber, metaDataName string,
+	metaData []byte, elements []*CoalesceElement) (
+	attrChangeTime time.Time, modificationTime time.Time, numWrites uint64, fileSize uint64, err error) {
 
 	var (
 		alreadyInInodeMap                  bool
+		coalesceTime                       time.Time
 		destInode                          *inMemoryInodeStruct
 		destInodeExtentMap                 sortedmap.BPlusTree
 		destInodeOffsetBeforeElementAppend uint64
@@ -1205,6 +1206,8 @@ func (vS *volumeStruct) Coalesce(destInodeNumber InodeNumber, metaDataName strin
 	// collect the NumberOfWrites value while locked (important for Etag)
 	numWrites = destInode.NumWrites
 	fileSize = destInode.Size
+	attrChangeTime = destInode.AttrChangeTime
+	modificationTime = destInode.ModificationTime
 
 	// Now, destInode is fully assembled... need to remove all elements references to currently shared LogSegments
 
--- a/pfs_middleware/pfs_middleware/middleware.py
+++ b/pfs_middleware/pfs_middleware/middleware.py
@@ -475,12 +475,47 @@ def mung_etags(obj_metadata, etag, num_writes):
     return
 
 
-def unmung_etags_and_return_best(obj_metadata, account_name, inum, num_writes,
-                                 is_dir=False, container_listing=False):
-    '''Unmung the ETag headers associated with an object to return
-    them to the state they were in when passed to pfs_middleware or
-    delete them if the object has changed.
+def unmung_etags(obj_metadata, num_writes):
+    '''
+    Unmung the ETag headers associated with an object to return
+    them to the state they were in when passed to pfs_middleware.
 
+    Delete them if the object has changed or the header value
+    does not parse correctly.
+    '''
+
+    # if the header is invalid or stale it is not added back after the pop
+    if LISTING_ETAG_OVERRIDE_HEADER in obj_metadata:
+        val = obj_metadata.pop(LISTING_ETAG_OVERRIDE_HEADER)
+        try:
+            stored_num_writes, rest = val.split(':', 1)
+            if int(stored_num_writes) == num_writes:
+                obj_metadata[LISTING_ETAG_OVERRIDE_HEADER] = rest
+        except ValueError:
+            pass
+
+    if S3API_ETAG_HEADER in obj_metadata:
+        val = obj_metadata.pop(S3API_ETAG_HEADER)
+        try:
+            stored_num_writes, rest = val.split(':', 1)
+            if int(stored_num_writes) == num_writes:
+                obj_metadata[S3API_ETAG_HEADER] = rest
+        except ValueError:
+            pass
+
+    if ORIGINAL_MD5_HEADER in obj_metadata:
+        val = obj_metadata.pop(ORIGINAL_MD5_HEADER)
+        try:
+            stored_num_writes, rest = val.split(':', 1)
+            if int(stored_num_writes) == num_writes:
+                obj_metadata[ORIGINAL_MD5_HEADER] = rest
+        except ValueError:
+            pass
+
+
+def best_possible_etag(obj_metadata, account_name, inum, num_writes,
+                       is_dir=False, container_listing=False):
+    '''
     Return the ETag that is most likely to be correct for the query,
     but leave other valid ETags values in the metadata, in case a
     higher layer filter wants to use them to override the value
@@ -490,47 +525,6 @@ def unmung_etags_and_return_best(obj_metadata, account_name, inum, num_writes,
     new ProxyFS ETag based on the account name, inode number, and
     number of writes.
     '''
-
-    # purge headers made invalid by subsequent writes/truncates from
-    # the returned headers
-    if LISTING_ETAG_OVERRIDE_HEADER in obj_metadata:
-        val = obj_metadata[LISTING_ETAG_OVERRIDE_HEADER]
-        try:
-            stored_num_writes, rest = val.split(':', 1)
-            if int(stored_num_writes) == num_writes:
-                obj_metadata[LISTING_ETAG_OVERRIDE_HEADER] = rest
-            else:
-                del obj_metadata[LISTING_ETAG_OVERRIDE_HEADER]
-        except ValueError:
-            del obj_metadata[LISTING_ETAG_OVERRIDE_HEADER]
-            pass
-
-    if S3API_ETAG_HEADER in obj_metadata:
-        val = obj_metadata[S3API_ETAG_HEADER]
-        try:
-            stored_num_writes, rest = val.split(':', 1)
-            if int(stored_num_writes) == num_writes:
-                obj_metadata[S3API_ETAG_HEADER] = rest
-            else:
-                del obj_metadata[S3API_ETAG_HEADER]
-        except ValueError:
-            del obj_metadata[S3API_ETAG_HEADER]
-            pass
-
-    if ORIGINAL_MD5_HEADER in obj_metadata:
-        val = obj_metadata[ORIGINAL_MD5_HEADER]
-
-        try:
-            stored_num_writes, rest = val.split(':', 1)
-            if int(stored_num_writes) == num_writes:
-                obj_metadata[ORIGINAL_MD5_HEADER] = rest
-            else:
-                del obj_metadata[ORIGINAL_MD5_HEADER]
-        except ValueError:
-            del obj_metadata[ORIGINAL_MD5_HEADER]
-            pass
-
-    # decide the best possible ETag to return
     if is_dir:
         return EMPTY_OBJECT_ETAG
 
@@ -1407,6 +1401,8 @@ class PfsMiddleware(object):
             last_modified = iso_timestamp_from_epoch_ns(ent.get(
                 "AttrChangeTime", ent["ModificationTime"]))
             obj_metadata = deserialize_metadata(ent["Metadata"])
+            unmung_etags(obj_metadata, ent["NumWrites"])
+
             content_type = swift_code.wsgi_to_str(
                 obj_metadata.get("Content-Type"))
             if content_type is None:
@@ -1415,10 +1411,10 @@ class PfsMiddleware(object):
             content_type, swift_bytes = content_type.partition(
                 ';swift_bytes=')[::2]
 
-            etag = unmung_etags_and_return_best(
-                obj_metadata,
-                account_name, ent["InodeNumber"], ent["NumWrites"],
-                is_dir=ent["IsDir"], container_listing=True)
+            etag = best_possible_etag(
+                obj_metadata, account_name,
+                ent["InodeNumber"], ent["NumWrites"], is_dir=ent["IsDir"],
+                container_listing=True)
             json_entry = {
                 "name": name,
                 "bytes": int(swift_bytes or size),
@@ -1443,6 +1439,8 @@ class PfsMiddleware(object):
         for container_entry in container_entries:
             obj_name = container_entry['Basename']
             obj_metadata = deserialize_metadata(container_entry["Metadata"])
+            unmung_etags(obj_metadata, container_entry["NumWrites"])
+
             content_type = swift_code.wsgi_to_str(
                 obj_metadata.get("Content-Type"))
             if content_type is None:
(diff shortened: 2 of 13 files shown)


pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -438,30 +439,111 @@ def extract_container_metadata_from_headers(req):
...
         try:
             stored_num_writes, rest = val.split(':', 1)
             if int(stored_num_writes) == num_writes:
-                return rest
+                obj_metadata[LISTING_ETAG_OVERRIDE_HEADER] = rest
+            else:
+                del obj_metadata[LISTING_ETAG_OVERRIDE_HEADER]
         except ValueError:
+            del obj_metadata[LISTING_ETAG_OVERRIDE_HEADER]
             pass
```
tipabu, line 521: nit: The pass is unnecessary now that we're doing something in the `except:` block.

I wonder if we could bring it closer to the original a bit, though; something like
```
    if LISTING_ETAG_OVERRIDE_HEADER in obj_metadata:
        val = obj_metadata.pop(LISTING_ETAG_OVERRIDE_HEADER)
        try:
            stored_num_writes, rest = val.split(':', 1)
            if int(stored_num_writes) == num_writes:
                obj_metadata[LISTING_ETAG_OVERRIDE_HEADER] = rest
        except ValueError:
            pass
```

and similar for the other headers.

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -438,30 +439,111 @@ def extract_container_metadata_from_headers(req):
...
+        obj_metadata[S3API_ETAG_HEADER] = "%d:%s" % (
+            num_writes, obj_metadata[S3API_ETAG_HEADER])
+
+    if etag is not None:
+        obj_metadata[ORIGINAL_MD5_HEADER] = "%d:%s" % (num_writes, etag)
+
+    return
+
+
+def unmung_etags_and_return_best(obj_metadata, account_name, inum, num_writes,
```
tipabu: Should this should be broken up into two functions?
```
def unmung_etags(obj_metadata, num_writes):
```
which _only_ seeks to do the inverse of `mung_etags()`, and
```
def best_possible_etag(obj_metadata, account_name, inum,
                       is_dir=False, container_listing=False):
```
which expects `unmung_etags()` to have already been called and no longer needs `num_writes`...

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -1471,9 +1554,11 @@ def put_object_as_directory(self, ctx):
             rpc_resp)
 
         resp_headers = {
-            "Etag": construct_etag(ctx.account_name, inode, num_writes),
             "Content-Type": DIRECTORY_CONTENT_TYPE,
             "Last-Modified": last_modified_from_epoch_ns(mtime_ns)}
+        resp_headers["Etag"] = unmung_etags_and_return_best(
```
tipabu: Huh. Why isn't this _always_ `EMPTY_OBJECT_ETAG`? When we're making a directory, it's just a zero-byte object from Swift's perspective, right? But this was already kinda funky...

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -1905,10 +1987,48 @@ def coalesce_object(self, ctx, auth_cb):
...
+            return err
+
+        # retrieve the ETag (or etag) value in the request (or None)
+        req_etag = req.headers.get('Etag', None)
+        if req_etag is None:
+            req_etag = req.headers.get('ETag', None)
+        if req_etag is None:
+            req_etag = req.headers.get('ETAG', None)
+        if req_etag is None:
+            req_etag = req.headers.get('etag', None)
```
tipabu: `req` is a `swob.Request`, right? All of these should be equivalent since `req.headers` is a [`HeaderEnvironProxy`](https://github.com/openstack/swift/blob/2.21.0/swift/common/swob.py#L236-L281) which translates them all into

    req.environ.get('HTTP_ETAG')

(Also, the `, None` should be unnecessary since we're using `.get()` where `None` is automatically the default.)

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -1905,10 +1987,48 @@ def coalesce_object(self, ctx, auth_cb):
                         return denial_response
                 authed_containers.add(elem_container_path)
 
+        # treat the number of objects as the number of writes
```
tipabu: It's not just that we're _treating it_ as such, right? It's that proxyfs **will** be using the number of coalesced files as the number of writes.

inode/file.go
```diff
@@ -1187,16 +1193,21 @@ func (vS *volumeStruct) Coalesce(destInodeNumber InodeNumber, elements []*Coales
...
-
 	destInode.CreationTime = coalesceTime
 	destInode.AttrChangeTime = coalesceTime
 	destInode.ModificationTime = coalesceTime
 
-	numWrites = destInode.NumWrites
+	// attach new middleware headers (why are we making a copy?)
+	inodeStreamBuf := make([]byte, len(metaData))
+	copy(inodeStreamBuf, metaData)
+	destInode.StreamMap[metaDataName] = inodeStreamBuf
```
tipabu, line 1204: This will all be fine when `metaData` is `nil`, right? Just thinking about rolling upgrades, when you may have an old proxy talking to an upgraded proxyfs...

pfs_middleware/tests/test_pfs_middleware.py
```diff
@@ -4072,6 +4136,26 @@ def mock_RpcCoalesce(coalese_req):
...
+                         '10340ab593ac8c32290a278e36d1f8df')
+        self.assertIn('X-Object-Sysmeta-S3Api-Acl', headers)
+        self.assertIn('X-Object-Sysmeta-S3Api-Etag', headers)
+        self.assertIn('X-Object-Sysmeta-Container-Update-Override-Etag',
+                      headers)
+        self.assertNotIn('X-Static-Large-Object', headers)
+        self.assertNotIn('X-Object-Sysmeta-Slo-Etag', headers)
+        self.assertNotIn('X-Object-Sysmeta-Slo-Size', headers)
+
+        print("test_success(): headers: '%s'\n" % (headers))
```
tipabu: Might want to take out the print.

pfs_middleware/tests/test_pfs_middleware.py
```diff
@@ -4438,33 +4522,33 @@ class TestBestPossibleEtag(unittest.TestCase):
...
-            mware.best_possible_etag(
+            mware.unmung_etags_and_return_best(
                 {self.HEADER: "1:5484c2634aa61c69fc02ef5400a61c94"},
                 "AUTH_test", 6676743, 1),
             '5484c2634aa61c69fc02ef5400a61c94')
 
     def test_modified(self):
         self.assertEqual(
-            mware.best_possible_etag(
+            mware.unmung_etags_and_return_best(
```
tipabu: Since this now has two responsibilities, we should probably keep a reference to the dict and check that `self.HEADER` is no longer present after the call.

tipabu (commented):
As a heads-up, I mostly looked at the pfs_middleware code, not so much the Go code. And I haven't actually functionally tested to verify all the plumbing.

I think I like the general approach, but
- I might prefer better separation of concerns with `unmung_etags_and_return_best()`,
- there's some redundancy when we go to get the COALESCE ETag, and
- we should probably have better testing of the unmunging of valid metadata and removal of invalid metadata.

We might also want to switch `aws-multipart-test.py` to use `boto3` directly, but that seems beyond the scope of what ought to happen before a merge.

pfs_middleware/pfs_middleware/middleware.py
charmster: good point!  i'm not that good at python ... i'll fix that.

pfs_middleware/pfs_middleware/middleware.py
charmster: well, yes and no!

formerly (before this patch) proxyfs tracked the number of writes for each object in the SLO and then added them together and returned that total to `parse_coalesce_object_response()`, which means it could have been different from the number of objects.

this patch also changes the proxyfs code to return (and store) the number of objects as the number of writes, instead of the total number of writes to create the component objects.

since the result of the COALESCE is a new object, i think we can assign any number of writes we want (including "1") and the ETag code will still work correctly.  so perhaps i should have changed this code, and corresonding proxyfs code, to make this `1`.

so i changed the way num_writes is calculated by proxyfs `Coalesce()`  to try and make the whole operation more transactional.  

if i don't know `num_writes` until `parse_coalesce_object_response()` returns, then i can't mung the metadata until then and i would have to perform a second transaction to set the correct metadata after the coalesce transaction, which opens the door (or window) to it never getting done (if the node crashes between updates).

so i took a sort of "middle of the road" approach my using the number of objects as the number of writes instead of the actual number of writes or simply using 1.

in my opinion, the comment is fine ... what do you think i should change the comment to say?

pfs_middleware/pfs_middleware/middleware.py
charmster: yes, it should probably be two functions.

i made this a single function to cut down the number of lines of code changed to make this change more palatable.

pretty much everywhere we call `best_possible_etag()` we have to call `unmung_etags()` first and there is no need to call `unmung_etags()` if you're not going to call `best_possible_etag()`.  so they are kind of "joined at the hip".

but i'm happy to change the code that looks like:
```            
          etag = unmung_etags_and_return_best(
                 obj_metadata, account_name,
                 container_entry["InodeNumber"],
                 container_entry["NumWrites"],
```
to:
```
            unmung_etags(obj_metadata)
            etag = best_possible_etag(
                 obj_metadata, account_name,
                 container_entry["InodeNumber"],
                 container_entry["NumWrites"],
```
i will do that.

pfs_middleware/pfs_middleware/middleware.py
charmster: sorry, i didn't address all of your comment!

anyway, if the conditional ETag header is no longer valid because somebody/something has written to the file, why do we want to keep it?  it just uses up disk space and network bandwidth.

is there a reason to keep `LISTING_ETAG_OVERRIDE_HEADER` once it is no longer germane?

tipabu: Are we talking about keeping it persisted in the metadata stream in proxyfs? I'd assumed because we didn't want to have to go digging through the metadata to figure out what's still valid and what isn't on every write.

Or are we talking about in the request path? We definitely don't need to keep it around then; the `.pop()` will do just that:

```
>>> x={'foo': 'bar'}
>>> x
{'foo': 'bar'}
>>> x.pop('foo')
'bar'
>>> x
{}
```
```

File: pfs_middleware/pfs_middleware/middleware.py (hunk 3 above)
Comment (charmster) [reply pfs_middleware/pfs_middleware/middleware.py]: the short answer is "i don't know."  the longer answer is "i don't know and i didn't really want to spend the time to figure it out."

from looking at the other code, i suspect that the PUT request that creates a directory could be coming from the `s3api` filter and it could supply its own value for the ETag, with the header name `S3API_ETAG_HEADER`.  there could also be a `LISTING_ETAG_OVERRIDE_HEADER`.

so, while a header named `ETag` (or `etag` or `Etag`) should have the value `EMPTY_OBJECT_ETAG`, there could also be a header named `S3API_ETAG_HEADER`, and if that header exists we should(?) return that ETag.

but, looking at the code for `unmung_etags_and_return_best()`, i can see that for a directory that function will only ever return `EMPTY_OBJECT_ETAG`.

i'm not sure if that's the correct behavior or not; i think i just copied what the earlier code seemed to do (see the above comment about "... and i don't really want to spend the time to figure it out.").

frankly, if the s3api had set a value for `LISTING_ETAG_OVERRIDE_HEADER` or `S3API_ETAG_HEADER` for the directory then i think `unmung_etags_and_return_best()` should return that ETag for the directory.

what do you think?

File: pfs_middleware/pfs_middleware/middleware.py (hunk 4 above)
Comment (charmster) [reply pfs_middleware/pfs_middleware/middleware.py]: i don't know anything about `swob.Request` and did not take the time to learn what it does (i was already having enough fun with the other inherited classes that seemed to affect how this code works).  i was fairly certain the code would work in all cases and not introduce any bugs.

from looking through the Swift code base, it seemed like there is a _diversity of opinion_ about capitalization:  `etag` could be lower-case (`etag`), upper-case (`ETAG`), title-case (`Etag`) or random case (`ETag`).  different filters actually seemed to convert the case of headers as the request passes through as they see fit!

so, if you're sure i can replace the existing code:
```
        # retrieve the ETag (or etag) value in the request (or None)                                                     
        req_etag = req.headers.get('Etag', None)
        if req_etag is None:
            req_etag = req.headers.get('ETag', None)
        if req_etag is None:
            req_etag = req.headers.get('ETAG', None)
        if req_etag is None:
            req_etag = req.headers.get('etag', None)
```
with
```
        # retrieve the ETag (or etag) value in the request (or None)                                                     
        req_etag = req.headers.get('Etag')
```
then i'm happy to make that change!

File: pfs_middleware/pfs_middleware/middleware.py (hunk 1 above)
Comment (charmster) [reply pfs_middleware/pfs_middleware/middleware.py:521]: Ahhh.  yes, the .pop() is better.

File: pfs_middleware/pfs_middleware/middleware.py (hunk 3 above)
Comment (charmster) [reply pfs_middleware/pfs_middleware/middleware.py]: i still think the correct behavior here is kind've murky.

_arguably_ if `S3API_ETAG_HEADER` was supplied with the request then that's what should be returned.  however, it is also true that the old `best_possible_etag()` code will return `EMPTY_OBJECT_ETAG` for all directories (or, more accurately, `EMPTY_OBJECT_ETAG` if the _optional_ `is_dir` is true; it default to false).

the original code seemed to have:
```
         resp_headers = {
-            "Etag": construct_etag(ctx.account_name, inode, num_writes),
             "Content-Type": DIRECTORY_CONTENT_TYPE,
             "Last-Modified": last_modified_from_epoch_ns(mtime_ns)}
```
which definitely seems wrong.

looking through the code, `get_object()` does not set the optional `is_dir` argument when calling `best_possible_etag()` and `get_object()` does not return the `IsDir` metadta..  however, the current behavior of proxyfsd is for a `get_object()` on a directory to fail with a `500` series error which seems quite wrong.

i have a fix for that (to fix problems that Burton is seeing) but it looks like that fix will need to percolate through the Python `get_object()` code.

anyway, in your opinion, should the ETag for a directory always be `EMPTY_OBJECT_ETAG` regardless of any headers set by the S3API layer?

Comment (charmster) [reply pfs_middleware/pfs_middleware/middleware.py]: i changed the code again on the theory that `best_possible_etag()` might change in the future.  so now it looks like:
```
        # currently best_possible_etag() returns EMPTY_OBJECT_ETAG for                                                   
        # all directories, but that might change in the future.                                                          
        # unmung the obj_metadata so best_possible_etag() can use it                                                     
        # if its behavior changes (note that num_writes is forced to 0).                                                 
        unmung_etags(obj_metadata, 0)
        resp_headers = {
            "Content-Type": DIRECTORY_CONTENT_TYPE,
            "Last-Modified": last_modified_from_epoch_ns(mtime_ns)}
        resp_headers["ETag"] = best_possible_etag(
            obj_metadata, ctx.account_name, inode, 0, is_dir=True)
```

File: pfs_middleware/pfs_middleware/middleware.py (hunk 4 above)
Comment (charmster) [reply pfs_middleware/pfs_middleware/middleware.py]: i dug into this some more, looked at `swob.Request` and discovered that it eventually overloads dictionaries (maps?) with its own key munger `def header_to_environ_key(header_name)`, which does what you said.

there are some things that, having been seen, cannot be unseen.

but i changed the code to match your suggestion.

File: pfs_middleware/pfs_middleware/middleware.py (hunk 5 above)
Comment (charmster) [reply pfs_middleware/pfs_middleware/middleware.py]: i updated the comment to:
```
        # proxyfs treats the number of objects as the number of writes                                                   
```

File: pfs_middleware/tests/test_pfs_middleware.py (hunk 7 above)
Comment (charmster) [reply pfs_middleware/tests/test_pfs_middleware.py]: yes, the `print()` is now gone.

File: pfs_middleware/tests/test_pfs_middleware.py (hunk 8 above)
Comment (charmster) [reply pfs_middleware/tests/test_pfs_middleware.py]: i had to rewrite that code again and found i had to copy() dictionaries to get it to work right.  i think that addresses this comment ...

File: inode/file.go (hunk 6 above)
Comment (charmster) [reply inode/file.go:1204]: Good catch!

curiously, while an uninitialized  slice is nil, taking the length of it still returns 0.

>  The zero value of a slice is nil. The len and cap functions will both return 0 for a nil slice. 

from https://blog.golang.org/go-slices-usage-and-internals

i tried it out in the Go playground, including using the gob encoder/decoder to see how it did with fields that are not sent and it seems fine.

https://play.golang.org/p/Qgtd_Ai6cD1

File: pfs_middleware/pfs_middleware/middleware.py (hunk 2 above)
Comment (charmster) [reply pfs_middleware/pfs_middleware/middleware.py]: making this change turned out to be more complicated and, arguably, fixed a bug in POST and perhaps bugs where metadata was getting modified when it shouldn't have been (because i didn't make a copy of the dictionary when i probably should have).

**Comment by charmster** [comment]
@tipabu can i get you take a look at my last set of changes and see if all of your concerns are resolved?

the two most controversial behaviors are probably this code in POST (post_object()) which allows certain ETAG values to be set if they were not set previously (which, i believe, is the previous behavior):

https://github.com/swiftstack/ProxyFS/blob/9174a329a59d70f76d1df0e486ebc74cc91e53e8/pfs_middleware/pfs_middleware/middleware.py#L1724-L1743

and these comments about the way ETAGs are contstructed which, again, i think matches existing behavior:
https://github.com/swiftstack/ProxyFS/blob/9174a329a59d70f76d1df0e486ebc74cc91e53e8/pfs_middleware/tests/test_pfs_middleware.py#L4569-L4581

but which leads to the need for this code:
https://github.com/swiftstack/ProxyFS/blob/9174a329a59d70f76d1df0e486ebc74cc91e53e8/pfs_middleware/pfs_middleware/middleware.py#L1736-L1739

File: pfs_middleware/pfs_middleware/middleware.py [hunk 9]
```diff
@@ -1462,18 +1543,29 @@ def put_object_as_directory(self, ctx):
             return swob.HTTPUnprocessableEntity(request=req)
 
         path = urllib_parse.unquote(req.path)
-        obj_metadata = serialize_metadata(extract_object_metadata_from_headers(
-            req.headers))
+        obj_metadata = extract_object_metadata_from_headers(req.headers)
+
+        # mung the passed etags, if any (NumWrites for a directory is
+        # always zero)
+        mung_etags(obj_metadata, request_etag, 0)
```
tipabu, line 1582: :+1: good call

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -1635,8 +1721,22 @@ def post_object(self, ctx):
             else:
                 raise
 
+        # We need to unmung and then mung the headers because the POST may
+        # supply a value S3API_ETAG_HEADER or LISTING_ETAG_OVERRIDE_HEADER
+        # (or ORIGINAL_MD5_HEADER?).  While merge_object_metadata() will
+        # not let these overwrite an existing header, they can be added to
+        # an object that doesn't have such a header.
```
tipabu: Hmmm.... we shouldn't be adding any new sysmeta on a POST.... but it looks like this is pre-existing behavior. We should maybe fix it at some point, but doesn't have to be part of this.

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -1635,8 +1721,22 @@ def post_object(self, ctx):
...
+        #
+        # Note that if an ETag header is stale (because NumWrites has
+        # changed) then unmung_etags() will strip it and it will not be
+        # put back by mung_etags().
```
tipabu: Yeah, that should be fine. Since the underlying data changed, we have no use for it any more.

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -1829,6 +1932,7 @@ def head_object(self, ctx):
             rpc.parse_head_response(head_response)
 
         headers = swob.HeaderKeyDict(deserialize_metadata(raw_md))
+        unmung_etags(headers, num_writes)
```
tipabu: This works equally-well with both `dict`s and `HeaderKeyDict`s, right?

pfs_middleware/pfs_middleware/middleware.py
tipabu: > there are some things that, having been seen, cannot be unseen.

lol! You should see what I'm having to do for py3... or maybe not, as the case may be.

FWIW, the opinions about ETag casing [get even stronger in s3api](https://github.com/openstack/swift/blob/2.21.0/swift/common/middleware/s3api/s3response.py#L34-L36). Some clients just don't understand that headers are supposed to be case-insensitive :frowning_face:

tipabu (approved):

pfs_middleware/pfs_middleware/middleware.py
charmster: OK, i wasn't sure of the correct semantics.

i will leave this as existing behavior.

pfs_middleware/pfs_middleware/middleware.py
charmster: i have no idea!  i will need to investigate ...

tipabu (approved):

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -402,7 +417,13 @@ def extract_object_metadata_from_headers(headers):
         if (header.startswith("X-Object-Meta-") or
                 header.startswith("X-Object-Sysmeta-") or
                 header in SPECIAL_OBJECT_METADATA_HEADERS):
-            meta_headers[header] = value
+
+            # do not let a client pass in ORIGINAL_MD5_HEADER
+            if (header != ORIGINAL_MD5_HEADER and
+                header != ORIGINAL_MD5_HEADER.title()):
```
tipabu: I think `tox -e lint` will be happier with something like
```
if header != ORIGINAL_MD5_HEADER and \
        header != ORIGINAL_MD5_HEADER.title():
    meta_headers[header] = value
```
though maybe it'd be more pythonic as
```
if header not in (ORIGINAL_MD5_HEADER,
                  ORIGINAL_MD5_HEADER.title()):
    meta_headers[header] = value
```
? The reduced repetition might even let it all fit on one line.

PR #309, opened 2019-06-07
Make pfs_middleware work with ss-release-2.21.0.4

clayg:


saio/bin/start_proxyfs_and_swift
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -x
```
edmc-ss: Not sure that you wanted to check this in. Kinda nice to keep this script quieter imho.

saio/Vagrantfile
```diff
@@ -14,6 +14,7 @@
 #     will collide in the contents of the $GOPATH/bin (only executables, not libraries)
 
 Vagrant.configure(2) do |config|
+  config.ssh.forward_agent = true
```
edmc-ss: What does this do btw? Does this make it so I don't have to do:
`vagrant ssh -- -Y` to get my X Display forwarded or something?

edmc-ss: Oh... I see... this makes it so I don't have to do:
`vagrant ssh -- -A` to get my .ssh creds over there... Cool!

edmc-ss: Perhaps we also need this line then:
`config.ssh.private_key_path = "~/.ssh/id_rsa"`, hmm?

edmc-ss: Sorry to be so anal about this.. the web sure does have a lot of discussion about this stuff... all new to me... particularly w.r.t. the host being OSX (as it is for me). I can try out your branch on Monday if you'd like...

PR #310, opened 2019-06-11
Start to document the saio setup a bit

clayg:
@edmc-ss I'd be happy to iterate on this with you outside of the critical pfs_middleware fixes!

### DIFF
--- a/saio/bin/start_and_mount_pfs
+++ b/saio/bin/start_and_mount_pfs
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -x
+
 # A simple script to start the services and mount the sample mount point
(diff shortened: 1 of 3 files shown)


saio/bin/start_and_mount_pfs
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -x
```
edmc-ss, line 3: Still do not prefer this in there to be honest...

edmc-ss (changes requested):
Just hoping to remove the `set -x` in `start_and_mount_pfs`...
I don't want to have to scroll thru a bunch of text to see if things worked as expected...but that's just me...

PR #349, opened 2019-07-31
Vg conf

edmc-ss:


confgen/api.go
```diff
@@ -0,0 +1,34 @@
+// Package confgen provides a mechanism by which a supplied configuration is converted into a
+// set of configuration files to be supplied to ProxyFS (proxyfsd), Samba (smbd et. al.), and
+// NFS (nfsd). In (at least) the case of Samba, multiple configuration files will be produced
+// with a per-IP path.
```
bschatz-swift, line 4: NIT: per VIP path

confgen/api.go
```diff
@@ -0,0 +1,34 @@
...
+//
+// Only the configuration files for the Cluster.WhoAmI-identified Peer will be produced though,
+// in the case of the ProxyFS configuration file, all VolumeGroups will be included given that
+// ProxyFS needs to know the mapping of VolumeGroup to Peer. The Cluster.WhoAmI can either already
+// be present in the `confFilePath` or be provided as an element of the `confOverrides` argument
+// in each of the APIs below.
```
bschatz-swift: I think we should state why a given proxyfsd needs to know the location of all VGs and their mapping to peer.    The real reason is bimodal middleware asks proxyfsd where a given volume is mounted.

confgen/api.go
```diff
@@ -0,0 +1,34 @@
...
+// As would be typical in a deployment where the placement of VolumeGroups on Peers in dynamic,
+// the `confFilePath` would provide `VirtualIPAddr` values as opposed to fixed `PrimaryPeer`
+// assignments. It is the callers responsibility to compute the assignments of VolumeGroups
+// to Peers and either modify the provided `confFilePath` to include PrimaryPeer values or to
+// supply those via the `confOverrides` (e.g. "VolumeGroup:CommonVolumeGroup.PrimaryPeer=Peer0").
+package confgen
+
+// ComputeInitial takes a supplied ConfFile, overlays ConfOverrides, and computes an initial
+// set of configuration files that are used by a per-IPAddr set of Samba instances as well as
+// NFSd & ProxyFS.
```
bschatz-swift, line 75: How does ss-proxyfs-monitord pass the mapping of VGs -> nodes?
Is that in the confOverrides array?
Wouldn't it be easier to pass this as a map?   How do you consume this?

confgen/api.go
edmc-ss: You know...that's what an earlier version actually said :-).
But I changed it to just "IP" to generically refer to both the VIP and the PublicIPAddr case (the latter being when the VIP field was left blank).
Make sense?

confgen/api.go
edmc-ss: I'll add that - thanks!

confgen/api.go
edmc-ss: I think the comment at the top kind of gives an example, no?
You would include in the confOverrides []string a string like "VolumeGroup:whatever.PrimaryPeer=PeerWhateverGUID".

edmc-ss: As for it being a map, that's true for the API but not for the cmdline version. Not sure how better to serialize a "map" for the cmdline :-)

edmc-ss: I should also say that this same mechanism is how the Cluster.WhoAmI is set as well...

confgen/api.go
bschatz-swift: ok but I don't think confgen should be used in the PubicIPAddr case.

confgen/api.go
bschatz-swift: ok

bschatz-swift (approved):
Time to merge this

PR #353, opened 2019-08-07
Bump Swift version in SAIO

gerardgine:


saio/vagrant_provision.sh
```diff
@@ -243,7 +243,7 @@ echo "export ST_KEY=testing" >> ~vagrant/.bash_profile
 cd ~swift
 git clone https://github.com/swiftstack/swift.git
 cd swift
-git checkout ss-release-2.21.0.4
+git checkout ss-release-2.22.0.2
```
clayg, line 246: Can this just pull the sha out of `pfs_middleware/tox.ini`?

gerardgine: Maybe? That's Ed's dev environment, and I'm just updating it as a courtesy. I agree, though, that the Swift version you be, at the very least, a little more "centralized". I'll try to work on that, but for now, I'll just merge this.

PR #391, opened 2019-10-18
Rework ChunkedPut retry to not keep connectionStruct if tcpConn fails

edmc-ss:


swiftclient/object.go
```diff
@@ -1401,11 +1405,14 @@ func (chunkedPutContext *chunkedPutContextStruct) retry() (err error) {
...
+	}
+	chunkedPutContext.connection, err = acquireChunkedConnection(chunkedPutContext.useReserveForVolumeName)
+	if nil != err {
 		chunkedPutContext.Unlock()
 		err = blunder.AddError(err, blunder.BadHTTPPutError)
-		logger.WarnfWithError(err, "swiftclient.chunkedPutContext.retry(\"%v/%v/%v\") openConnection() failed",
+		logger.WarnfWithError(err, "swiftclient.chunkedPutContext.retry(\"%v/%v/%v\") acquireChunkedConnection() failed",
 			chunkedPutContext.accountName, chunkedPutContext.containerName, chunkedPutContext.objectName)
 		return
 	}
```
charmster, line 1418: This code change affects the timing of retries in ways that we don't want (i think).

openConnection() will immediately attempt to open a connecction and succeed or fail (though it could take awhile to fail if the noauth proxy is down because it would have to timeout).

acquireChunkedConnection(), on the other hand, can can also block for arbitrary periods of time if all of the chunked put connections are in use. remember that, by design, when connections are in short supply a caller to acquireChunkedConnect() goes at the end of the line, behind everyone else waiting.

let's say that things are really busy, but making progress, so that it takes 60 sec to get a chunked put connection.  that means that every retry of the chunked PUT is going to take at least 60 sec, which is not the way the retry mechanism was designed to work. (to a certain extent the retry mechanism compensates by the way it computes the time to delay.  it measures from the start time of the last request to the start time of the next one, so once the retry delay goes over 60 sec this is not so important).

but i believe we discussed this at the time and concluded a request only needed to "wait in line" once to get a connection, and after that it didn't have to wait again (which is why the code you removed called `openConnection()` directly.

there's also a more subtle issue with this change.  if we get in a situation where the noauth proxy is overwhelmed, it will start shedding load by failing some requests early.  in that situation, the correct approach is to back off the rate at which we are sending requests.  this code change will have the effect of reducing the amount of "back off".

suppose the proxy server is overwhelmed and returning early errors, that we 100 routines wanting to do chunked PUTs and we have 20 connections in the ChunkedConnection pool.  with the old code, the first 20 goroutines will grab ChunkedConnections, send their requests, receive a failure, and back off (initial back-off 1 sec).  after a 1 sec delay, they call openConnection() to get a new connection and try again.  if they get the failure they will then delay for 2 sec, wait and try again.  if the noauth proxy is still overwhelmed, they will delay for 4 sec and then try again.

meanwhile the 80 other go-routines have not gotten a connection yet and are just waiting for the first 20 chunked PUT to complete successfully or hit the full time-out window.

with this change, the first 20 go-routines will send their request.  after it fails, they wait 1 sec and release their ChunkedConnection, and which point the next set of 20 go-routines will get connections and send their requests, and so on.  the next effect is the overwhelmed proxy server gets hit with 20 requests/sec for 5 sec instead of just for 1 sec.  that's _probably_ not what we want.

charmster (commented):
I don't think the change to use acquireChunkedPutConnection() for retries is a good idea because it will affect the retry delays in some unpredictable ways when there are a lot of active connections (see the extended comment around the code change).

i won't claim that the current retry behavior is correct, but it is predictable and it does have a certain amount of testing in the field.  we should probably think carefully before we change it.

(this comment added later: i notice that most other swiftclient operations, like `objectGet()`, call `acquireNonChunkedConnection()` at the beginning of each retry.  i think that that is undesirable behavior for the reasons listed above ... although they do release the connection as soon as an error occurs, before sleeping for the backoff period, so the behavior will be different than both the old chunkedPutConnection retry code *and* the proposed chunkedPutConnection retry code.  we probably want the behavior to be same for all type of Swift operations ...)

PR #394, opened 2019-10-31
Fix bugs in confgen's search for templates.

charmster:
confgen no longers adds "template/<filename>" to the possible
configuration directory and now just tries "<filename>" so it can find
files that happen to reside in, say, /opt/ss/var/lib/proxyfsd as opposed
to expecting /opt/ss/var/lib/proxyfsd/templates.

confgen now chooses the first directory where it can Stat() the
`smb_globals.tmpl` file, instead of the first directory where it can
find and successfully parse such a file.

if, for some reason, we have a template file that can't be parsed this
will make for clearer error messages.  with the old code it would just
say that the file can't be found.  with this change "can't be found"
and "can't be parsed" will generate separate errors.


charmster:
The brilliance of this fix staggers the imagination.  It is truly an inspired work of genius.

Who could stand in the way of merging this masterpiece?

PR #396, opened 2019-11-05
Added pfs_middleware conf data to /info resource

edmc-ss:


pfs_middleware/pfs_middleware/__init__.py
```diff
@@ -16,12 +16,18 @@
...
 
 
 def filter_factory(global_conf, **local_conf):
     conf = global_conf.copy()
     conf.update(local_conf)
 
+    register_swift_info('pfs',
+                        proxyfsd_host=conf["proxyfsd_host"],
+                        proxyfsd_port=conf["proxyfsd_port"],
+                        bypass_mode=conf["bypass_mode"])
```
tipabu, line 29: If it were me, I'd only have exposed the bypass mode. proxyfsd host and port are internal to the cluster and shouldn't be directly usable by external clients, correct? The only information that should be useful to an external client would be that

* ProxyFS is enabled (or may be enabled) for at least some accounts in the cluster, and
* clients can use the PROXYFS verb.

PR #404, opened 2019-11-26
Bypass rpc

bschatz-swift:
Bypass RPC which tunnels metadata operations to proxyfsd.

Eventually, this RPC will be "hardended" to handle lost connections and retries as well as TLS

### DIFF
--- a/jrpcfs/api.go
+++ b/jrpcfs/api.go
@@ -165,6 +165,23 @@ import (
 type MountIDAsByteArray [16]byte
 type MountIDAsString string
 
+// BypassReply contains response to BypassReq
+type BypassReply struct {
+	// TODO - update all error vs result vs, etc
+	// want to include Result and Error???
+	ID     uint64      `json:"id"`
+	Result interface{} `json:"result"`
+	Error  string      `json:"error"`
+}
+
+// BypassReq tunnels file system RPCs from pfsagentd to proxyfsd
+type BypassReq struct {
+	JSONrpc string         `json:"jsonrpc"`
+	Method  string         `json:"method"`
+	ID      uint64         `json:"id"`
+	Params  [1]interface{} `json:"params"`
+}
+
 // InodeHandle is embedded in a number of the request objects.
 type InodeHandle struct {
 	MountID     MountIDAsString
--- a/jrpcfs/api.go
+++ b/jrpcfs/api.go
@@ -176,10 +176,10 @@ type BypassReply struct {
 
 // BypassReq tunnels file system RPCs from pfsagentd to proxyfsd
 type BypassReq struct {
-	JSONrpc string         `json:"jsonrpc"`
-	Method  string         `json:"method"`
-	ID      uint64         `json:"id"`
-	Params  [1]interface{} `json:"params"`
+	JSONrpc string `json:"jsonrpc"`
+	Method  string `json:"method"`
+	ID      uint64 `json:"id"`
+	Params  []byte `json:"params"`
 }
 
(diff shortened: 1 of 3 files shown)


jrpcfs/api.go
```diff
@@ -165,6 +165,23 @@ import (
...
+	Result []byte `json:"result"`
+	Error  string `json:"error"`
+}
+
+// BypassReq tunnels file system RPCs from pfsagentd to proxyfsd
+type BypassReq struct {
+	JSONrpc string `json:"jsonrpc"`
+	Method  string `json:"method"`
+	ID      uint64 `json:"id"`
+	Params  []byte
```
edmc-ss: So I see that you just put a byte slice here... I guess that means you plan on un-marshaling the payload of Params separately? That's just fine... though I didn't do it that way. You should be able to see how I did it... I basically un-marshaled just the things I needed at this point (in your case, you need Method only I should think)... then, once you know the Method, you'd know how to un-marshal what you find in Params []byte. That's fine, too...

As for the line, though, you'll still need to "adorn" it with `json:"params"` since that's what the field will be named "on the wire" (i.e. in JSON-RPC format).

jrpcfs/api.go
```diff
@@ -165,6 +165,23 @@ import (
 type MountIDAsByteArray [16]byte
 type MountIDAsString string
 
+// BypassReply contains response to BypassReq
+type BypassReply struct {
+	// TODO - update all error vs result vs, etc
+	// want to include Result and Error???
+	ID     uint64 `json:"id"`
+	Result []byte `json:"result"`
```
edmc-ss: See my similar comment... you "could" do what I did and put an `interface{}` here and, at time of marshaling, assign the field the particular type of the return from the method call you make.

jrpcfs/bypass_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+// TODO - move to API???
+func bypassMarshalReq(requestMethod string, ID uint64, tunnelReq interface{}) (bypassReq *BypassReq, err error) {
+
+	// Create request struct and convert tunnelReq to JSON
+	// TODO - need mount ID???
+	bypassReq = &BypassReq{
+		JSONrpc: "2.0",
+		Method:  requestMethod,
+		ID:      ID,
```
edmc-ss, line 24: Well this explains why you didn't need Params to be "json-annotated" (see earlier comment) :-)

BTW, I should have mentioned, there is actually an RpcPing API that would be most handy for your test code to actually exercise... I believe it responds with the # of bytes in whatever is sent to RpcPing()...

jrpcfs/bypass_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	assert := assert.New(t)
+
+	// Create a RpcPing request
+	pingReq := &PingReq{
+		Message: "TestMessage",
+	}
+
+	// Create bypass request
+	bypassReq, marshalErr := bypassMarshalReq("Server.RpcPing", dummyRequestID, pingReq)
```
edmc-ss, line 46: Ah... here you used RpcPing... excellent!

edmc-ss (commented):
This looks very good... love to see it when the test runs successfully :-). Does it?

You might check out the tricks I did in pfsagentd/jrpc.go & pfsagentd/requests.go to see if there is some convention for avoiding "nested" JSON marshal/un-marshal work on your part. It might be clearer...

bschatz-swift:
These comments have been incorporated in the latest source so deleting this branch

PR #406, opened 2019-12-04
retryRpc package to implement retryable RPCs with exactly once semantics

bschatz-swift:
api.go - API for package with comments to show algorithm and open issues
api_test.go - shows how the API would be used by the caller

### DIFF
--- a/retryrpc/api.go
+++ b/retryrpc/api.go
@@ -8,6 +8,7 @@
 package retryrpc
 
 import (
+	"reflect"
 	"sync"
 	"syscall"
 	"time"
@@ -44,7 +45,7 @@ type ReqResKey struct {
 type Server struct {
 	sync.Mutex
 	completedTTL      time.Duration // How long a completed request stays on queue
-	serviceMap        map[string]*func(request interface{}, response interface{})
+	serviceMap        map[string]*reflect.Method
 	pendingRequests   map[ReqResKey]*Request          // Request which have not been completed yet.
 	completedRequests map[ReqResKey]*completedRequest // Request which have been completed.  Will age out
 	ipaddr            string                          // IP address server listens too
@@ -54,7 +55,7 @@ type Server struct {
 // NewServer creates the Server object
 func NewServer(ttl time.Duration, ipaddr string, port int) *Server {
 	s := &Server{}
-	s.serviceMap = make(map[string]*func(request interface{}, response interface{}))
+	s.serviceMap = make(map[string]*reflect.Method)
 	s.completedTTL = ttl
 	s.ipaddr = ipaddr
 	s.port = port
@@ -62,11 +63,13 @@ func NewServer(ttl time.Duration, ipaddr string, port int) *Server {
 }
 
 // Register creates the map of server methods
-func (server *Server) Register() (err error) {
+func (server *Server) Register(retrySvr interface{}) (err error) {
 
+	// Find all the methods associated with retrySvr and put into serviceMap
 	// TODO - build the serviceMap similiar to how net/rpc Register does it...
 	// serviceMap looks like serviceMap["RpcPing"]*rpcPing()
-	return
+
+	return server.register(retrySvr)
 }
 
 // Run server loop, accept connections, read request, run RPC method and
--- a/retryrpc/api.go
+++ b/retryrpc/api.go
@@ -13,7 +13,7 @@ type Request struct {
 	Len       uint64 // Length of this struct including len field
 	RequestID uint64 // Unique ID of this request
 	MountID   uint64 // Mount ID
-	Req       []byte // JSON containing request
+	JReq      []byte // JSON containing request
 }
 
 // Response is the structure returned
@@ -87,12 +87,17 @@ func NewClient(mountID uint64) *Client {
 }
 
 // MakeRPC takes the RPC method and arguments and returns a Request struct
-func MakeRPC(method string, params []byte) (request *Request) {
+func MakeRPC(method string, args ...interface{}) (request *Request) {
+
+	// TODO - RequestID, MountID, JReq
+
 	// This will:
 	// 1. bump the client request ID
-	// 2. Store the RPC method in the request
-	// 3. Store the parameters in the request
-	// 4. return the request
+	// 2. Store the RPC method and args in JSON and store in JReq
+	// 3. Store the mountID
+	// 4. set the length of the request
+	// 5. return the request
+	request = &Request{}
 	return
 }
 
@@ -103,4 +108,10 @@ func (client *Client) Send(request *Request) (response *Response, err error) {
 	// off the queue
 	// if connection drops, get new connection and resend the request
 	return
+}
+
+// Close gracefully shuts down the client.   This allows the Server
+// to remove Client request on completed queue.
+func (client *Client) Close() {
+
 }
--- a/retryrpc/api.go
+++ b/retryrpc/api.go
@@ -18,43 +18,18 @@ import (
 	"github.com/swiftstack/ProxyFS/jrpcfs"
 )
 
-// Request is the structure sent
-type Request struct {
-	Len  int64  // Length of JReq
-	JReq []byte // JSON containing request
-}
-
-// Reply is the structure returned
-type Reply struct {
-	Len     int64  // Length of JResult
-	JResult []byte // JSON containing response - only valid if err != nil
-}
-
-type completedRequest struct {
-	completedTime time.Time
-	request       *Request
-}
-
-// ReqResKey uniquely identifies a client request from a client
-type ReqResKey struct {
-	uniqueID        string
-	clientRequestID uint64
-}
-
 // Server tracks the state of the server
 type Server struct {
 	sync.Mutex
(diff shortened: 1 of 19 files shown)


retryrpc/api.go
```diff
@@ -0,0 +1,134 @@
...
+package retryrpc
+
+import (
+	"container/list"
+	"sync"
+	"time"
+)
+
+// Request is the structure sent
+type Request struct {
```
edmc-ss: So by this type (Request) and the corresponding type Response, we are jettisoning actual JSON RPC, correct? Sure, we've decided to marshal via JSON... but we no longer need anything like the form that JSON RPC defines namely:
```
type jrpcRequestStruct struct {
	JSONrpc string         `json:"jsonrpc"`
	Method  string         `json:"method"`
	ID      uint64         `json:"id"`
	Params  [1]interface{} `json:"params"`
}

type jrpcResponseWithErrorStruct struct {
	ID     uint64      `json:"id"`
	Error  string      `json:"error"`
	Result interface{} `json:"result"`
}
```
Correct?

retryrpc/api.go
```diff
@@ -0,0 +1,134 @@
...
+	request       *Request
+}
+
+// Server tracks the state of the server
+type Server struct {
+	sync.Mutex
+	completedTTL      time.Duration // How long a completed request stays on queue
+	serviceMap        map[string]*func(request interface{}, response interface{})
+	connections       *list.List                   // List of outstanding connections
+	pendingRequests   map[string]*Request          // Request which have not been completed yet.
```
edmc-ss: For maps, I like to be told at least what the Key is... I can (usually) see what the Value is :-).

As for the Key being a `string`, I'm actually surprised at this. Perhaps you do something with RequestID and MountID to be the string that is the Key here? There are more direct ways to do that.

For instance, you can look at type `readCacheKeyStruct` in inode/config.go:
```
type readCacheKeyStruct struct {
	volumeName       string
	logSegmentNumber uint64
	cacheLineTag     uint64 // LogSegment offset / readCacheLineSize
}
```
It is used as the Key in the readCache field of the volumeGroupStruct listed just below.
In this case, since it's a named Key type, I didn't bother commenting what it is :-).

Anyway, this is just a cute thing you can do with Go maps I thought more appropriate here.

retryrpc/api.go
```diff
@@ -0,0 +1,134 @@
...
+	completedTime time.Time
+	request       *Request
+}
+
+// Server tracks the state of the server
+type Server struct {
+	sync.Mutex
+	completedTTL      time.Duration // How long a completed request stays on queue
+	serviceMap        map[string]*func(request interface{}, response interface{})
+	connections       *list.List                   // List of outstanding connections
```
edmc-ss: As per my comment just following, just saying *list.List fails to reference what the Values (Elements) actually are... a comment would be more helpful if it referenced a struct corresponding to the Connections that are on the list. You certainly need such a structure if only to wrap a net.Conn and the list.Element, right?

retryrpc/api.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// Run server loop, accept connections, read request, run RPC method and
+// return the results.
+func (server *Server) Run() {
+	// TODO - Algorithm - Standard server stuff
+	// 1. goroutine which accepts new connection and creates goroutine W
+	// 2. goroutine W adds request to server.pendingRequest, unmarshals
+	//    request, processes RPC, puts request on server.completedRequest
+	//    and under same lock removes from server.pendingRequest.
```
edmc-ss: I would suggest the "processes RPC" step etc... be done in a per-request Goroutine itself, right?

retryrpc/api.go
```diff
@@ -0,0 +1,134 @@
...
+
+// Close stops the server
+func (server *Server) Close() {
+	// TODO - blocks until the server is shut down
+
+}
+
+// Client methods
+
+// Client tracking structure
```
edmc-ss, line 97: So I gather this `retryrpc` package will be used by both package jrpcfs (or its replacement/peer) and program pfsagentd, correct? Sounds great!

retryrpc/api.go
```diff
@@ -0,0 +1,134 @@
...
+type Client struct {
+	currentRequestID uint64 // Last request ID - start from clock
+	// tick at mount and increment from there?
+	// Handle reset of time?
+	myUniqueID         uint64              // Unique ID across all clients
+	subsetID           uint64              // Subset of unique ID (mount ID)
```
edmc-ss: Re myUniqueID & subsetID... I don't think either make too much sense.
Don't we just want to have a []byte that holds whatever the client will be passing in the MountID field mentioned above? I was mostly confused why there are two such fields here...

retryrpc/api.go
```diff
@@ -0,0 +1,134 @@
...
+// Close stops the server
+func (server *Server) Close() {
+	// TODO - blocks until the server is shut down
+
+}
+
+// Client methods
+
+// Client tracking structure
+type Client struct {
```
edmc-ss, line 98: PFSAgent will only have one of these, correct?
I mean it does a single mount... and opens up a single connection.
I suppose other clients may want to manage multiple... presumably, they'd each have a unique connection to the server, right? We should probably make that clear by embedding in this structure things like the connection... Presumably, there'd also be a "shuttingDown" bool that is set when shutting down :-) and the Close() method sets before closing the network connection... thus the goroutine reading the connection would see the connection go down and check the "shuttingDown" bool to know that this was intentional (rather than re-connecting and re-transmitting each request).

retryrpc/api.go
bschatz-swift: Request.JReq will contain JSON with these fields:

    type jsonBypassRequest struct {
    	    Method  string         `json:"method"`
	    Params  [1]interface{} `json:"params"`
    }

and the Response.Res will contain JSON with these fields:

    typeBypassResponse struct {
             Result interface{}  `json:"result"`
    }

I will not be using jrpcRequestStruct or jrpcResponseWithErrorStruct.

Did that answer your question?

I will update the PR with these responses to include the marshaling and unmarshaling.

retryrpc/api.go
```diff
@@ -0,0 +1,134 @@
...
+	// 2. put request on client.outstandingRequests
+	// 3. Send request on socket to server and have goroutine block
+	//    on socket waiting for result
+	//    a. if read result then remove from queue and return result
+	//    b. if get error (which one?) on socket then resend request.
+	//       will have to make sure have enough info in request to make
+	//       the operation idempotent.   Assume client retries until
+	//       server comes back up?   Wait for failover to a peer?
+	//       Assume using VIP on proxyfs node?
+	// 4. Should we block forever? How kill?
```
edmc-ss: I think you need the Client.outstandingRequest map (hmm... do we need to re-transmit in time order and, hence, also have to maintain a time-ordered list of them?)... and to have a separate "Sender" goroutine. Similarly, there should be a "Receiver" goroutine that receives (out-of-order) responses looking them up in Client.outstandingRequest to know how to signal the Send() func that is blocked waiting on the completion, right? Thus, in Request, there should be something like a sync.WaitGroup and, perhaps, an error as well in case the RPC was "unsendable".

retryrpc/api.go
bschatz-swift: yes - that is correct

retryrpc/api.go
edmc-ss: Yeah, as I mentioned verbally, I doubt we need the whole JSON-RPC doc in there...

retryrpc/api.go
bschatz-swift: Good catch.   The key for pendingRequests will be a combination of mountID and client request ID.   I added this to the code:

// RequestKey uniquely identifies a client request from a client
type RequestKey struct {
	uniqueID string
	clientRequestID uint64
}

retryrpc/api.go
bschatz-swift: Not sure yet how I will manage the connections.   I think I will put it in the request that I queue on pendingRequest.   I will work on that.

retryrpc/api.go
bschatz-swift: Yes - updated comment

retryrpc/api.go
bschatz-swift: Good - I wanted to talk about this.

We said we wanted an opaque ID since we may break this off into a separate repo.  (That is why I did not name it mountID)

I guess in our case pfsagent will have a different mountID per mounted file system.   We do not have to worry about one pfsagent handling multiple mount points.   If we did, then we would have to have a unique ID for the client and a unique ID for the mount points it was using.

Therefore, myUniqueID is the only field we need.

retryrpc/api.go
bschatz-swift: Great!   That is what I wanted to clean up.    Yes - PFSAgent will only have one Client struct now that we are sure there is one PFSAgent per mount point.

I agree with the fields for shutting down, etc.   I am looking for feedback on the overall approach.

bschatz-swift:
The latest code incorporates these comments as needed so this branch can be deleted

PR #408, opened 2019-12-14
Stop thrashing memcache when using X-Bypass-ProxyFS

tipabu:


pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -838,7 +842,9 @@ def __call__(self, req):
                     pop_and_restore(req.environ, 'swift.authorize_override',
                                     False):
                 if auth_cb and req.environ.get('swift.source') != 'PFS':
-                    req.acl = self._fetch_appropriate_acl(ctx)
+                    if not is_bypass_request:
+                        req.acl = self._fetch_appropriate_acl(ctx)
+                    # else, user needs to be swift owner
```
tipabu, line 853: Even if _we_ don't call `get_container_info`, middlewares left of us likely **will**. And likely in a fresh pre-auth'ed env, so no `X-Bypass-ProxyFS` header... hmm... :thinking:

pfs_middleware/pfs_middleware/middleware.py
```diff
@@ -778,6 +778,12 @@ def __init__(self, app, conf, logger=None):
     def __call__(self, req):
         vrs, acc, con, obj = utils.parse_path(req.path)
 
+        # Alternate way to specify bypass mode: /proxyfs/AUTH_acct/...
+        if vrs == 'proxyfs':
+            req.headers['X-Bypass-ProxyFS'] = 'true'
```
tipabu, line 783: Should we continue to support using a header for bypass mode at all? As things stand, the upgrade path looks hairy -- old code that _only_ knew about the header will start getting a bunch of 404s on upgrade now that we've stopped clearing memcache.

If we drop the header, we could switch this all to be something more like

``` python
is_bypass_request = False
if vrs == 'proxyfs' and self.bypass_mode in ('read-only', 'read-write'):
    req.path_info = req.path_info.replace('/proxyfs/', '/v1/', 1)
    vrs = 'v1'
    is_bypass_request = True
```

PR #410, opened 2019-12-19
Improve PFSA logging

gerardgine:
Add volume name and PID to all messages
Remove redundant volume name from some messages

### DIFF
--- a/pfsagentd/log.go
+++ b/pfsagentd/log.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 	"os"
 	"time"
-
-	"golang.org/x/sys/unix"
 )
 
 func logFatal(err error) {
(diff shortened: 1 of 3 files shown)


gerardgine:
I'm not so sure about the change at https://github.com/swiftstack/ProxyFS/pull/410/files#diff-6240498d7b369ef3ab19bbe4dd922dedR26

As @edmc-ss mentioned, we're shipping this conf file as a template in the pfsagent package, and that log path is not consistent with the log paths we use in the controller.

Just defining `LogFilePath` as `pfsagentd.log` will make the log file be at `/pfsagentd.log` in Runway, which is not good either...

pfsagentd/log.go
```diff
@@ -42,8 +44,8 @@ func logf(level string, format string, args ...interface{}) {
 		logMsg         string
 	)
 
-	enhancedFormat = "%-32s %-5s " + format
-	enhancedArgs = append([]interface{}{time.Now().Format(time.RFC3339Nano), level}, args...)
+	enhancedFormat = "[%s][%s][%d][%s] " + format
+	enhancedArgs = append([]interface{}{time.Now().Format(time.RFC3339Nano), globals.config.FUSEVolumeName, unix.Getpid(), level}, args...)
```
edmc-ss: I'd prefer you call os.Getpid() rather than unix.Getpid(): https://golang.org/pkg/os/#Getpid

edmc-ss (changes requested):
After that change to use os.Getpid(), I'm good with this - thanks!

pfsagentd/log.go
gerardgine: Done! Any thoughts on the change to the config file?

PR #432, opened 2020-01-22
Fix NumWrites when doing inode.Wrote(...,patchOnly==true)

edmc-ss:


edmc-ss:
This change was incorrect... the code was already properly setting NumWrites

PR #442, opened 2020-02-10
pfs_middleware: stop requiring that bypass_mode config option be present

tipabu:


edmc-ss:
Incorporated as PR 442 (https://github.com/swiftstack/ProxyFS/pull/444)

PR #443, opened 2020-02-10
pfs_middleware: remove X-Bypass-ProxyFS header support

tipabu:
You need `/proxyfs/...` paths and new Swift to make things happy these days anyway.

While we're at it, make bimodal_checker aware of `/proxyfs/...` paths; how did this ever work before!?


edmc-ss:
Incorporated as PR 442 (https://github.com/swiftstack/ProxyFS/pull/444)

PR #446, opened 2020-02-13
Fix bugs in confgen and mkproxyfs.

charmster:
confen was emitting scripts to make /share directories that had syntax
errors so they wouldn't run (a missing closing `]` in an if statement.
annoyingly, running the script still exited with 0.

mkproxyfs was sending an error message to stdout so etcd-mgmt was not
picking it up.  change it to send the message to stderr.


charmster:
fix a bug in confgen that caused it to generate bad scripts to make directories under /share.

fix a very minor bug in mkproxyfs that sent the error message to stdout instead of stderr, which prevented etcd-mgmt from properly recording it.

PR #447, opened 2020-02-14
Fix another issue with confgen.

charmster:
the fuse_setup.bash script that it generates looks like:

set -e
if [ ! -d '/share/craigfs' ]; then
    mkdir -p -m 0000 '/share/craigfs'
fi

if '/share/craigfs' exists and has a file system mounted but proxyfs is
not running, the "-d '/share/craigfs'" will evaluate false so then the
mkdir will run.  the mkdir will fail with EEXIST and the script will abort
(due to "set -e").

about the only thing i can see to do is override the mkdir failure like so:

set -e
if [ ! -d '/share/craigfs' ]; then
    mkdir -p -m 0000 '/share/craigfs' || true
fi

Changed confgen to do that.

### DIFF
--- a/confgen/api_internal.go
+++ b/confgen/api_internal.go
@@ -293,9 +293,14 @@ func computeInitial(envMap EnvMap, confFilePath string, confOverrides []string, 
 
 	for _, volume = range localVolumeMap {
 		if "" != volume.FUSEMountPointName {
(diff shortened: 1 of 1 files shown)


edmc-ss:
@charmster : Should we hold off 1.15.3 tag (for SSController 7.3) for this?

PR #464, opened 2020-03-18
Convert bucketstats "bucket" Add() func to also keep an accurate runn…

edmc-ss:
…ing total

### DIFF
--- a/bucketstats/api.go
+++ b/bucketstats/api.go
@@ -216,12 +216,15 @@ func (this *Average) Sprint(stringFmt StatStringFormat, pkgName string, statsGro
 // values in bucket n is very slightly larger than 2^n.
 //
 type BucketLog2Round struct {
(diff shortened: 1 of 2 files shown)


edmc-ss:
BTW - I noticed that Averager does exactly this same trick (i.e. it uses atomicUint64Add() on both the count and the total fields... with no "mutex" that might make those two additions collectively atomic).

charmster:
As I made clear in multiple comments on Slack yesterday, where I also explained the cost tradeoffs, I disagree with increasing the CPU cost of adding to a bucketstat by 50 - 100% to get slightly more accurate averages and totals.

If you want to create a new type that does this, perhaps `BucketLog2RoundAccurateTotal`, that would be fine.  That would allow the user to make the relevant trade-off of CPU cost and accuracy.

But I'm going to roll back this change.

PR #465, opened 2020-03-19
Revert "Convert bucketstats "bucket" Add() func to also keep an accurate runn…"

charmster:
Reverts swiftstack/ProxyFS#464

### DIFF
--- a/bucketstats/api.go
+++ b/bucketstats/api.go
@@ -216,15 +216,12 @@ func (this *Average) Sprint(stringFmt StatStringFormat, pkgName string, statsGro
 // values in bucket n is very slightly larger than 2^n.
 //
 type BucketLog2Round struct {
-	total       uint64 // Ensure 64-bit alignment
 	Name        string
 	NBucket     uint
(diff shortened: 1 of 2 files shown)


charmster:
As I made clear in multiple comments on Slack yesterday, where I also explained the cost trade-offs, I disagree with increasing the CPU cost of adding to a bucketstat by 50 - 100% to get slightly more accurate averages and totals.

Therefore, I'm reverting this change.

If you want to create a new type that does this, perhaps BucketLog2RoundAccurateTotal, that would be fine. That would allow the user to make the relevant trade-off of CPU cost and accuracy depending on their needs.

You can use the same unit tests for `BucketLog2Round` to test `BucketLog2RoundAccurateTotal`, and just add another test that verifies the total and average are more accurate (the BucketLog2Round tests already verify its +/- 10%).

edmc-ss:
I guess we could converse here... but perhaps others not aware of this PR might miss it :-) so I responded in channel

PR #472, opened 2020-04-14
Idle socket handling

bschatz-swift:
Don't retransmit if idle, KEEPALIVE and fix empty header

    * fixed case where server was not setting header in request
    * send KEEPALIVE from client and server
    * if getIO times out and we have no outstanding requests - don't call
    retransmit


bschatz-swift:
abandoned

PR #473, opened 2020-04-15
Fix races in TestLocks

bschatz-swift:
Fix races in TestLocks as found by the go race detector

* instead of sleeping and assuming a lock will or will not be granted,
  change to use the existing WGs for force a lock to a certain state
* to force races between shared and exclusive goroutines, use a
goroutine to sleep and then release the lock


bschatz-swift:
no longer needed

PR #494, opened 2020-05-16
Upgrade jQuery due to CVEs reported by GitHub

gerardgine:
2 things remain to be solved:

- If file `jquery_underline_3_dot_2_dot_1_dot_min_dot_js_.go` is present from a previous build, the next build will fail. I guess we have to remove it if present at build time?
- When previous point is done, remove `jquery_underline_3_dot_2_dot_1_dot_min_dot_js_.go` from `httpserver/.gitignore`. If done earlier, we risk somebody comitting it by mistake.


httpserver/Makefile
```diff
@@ -4,6 +4,7 @@ generatedfiles := \
 	bootstrap_dot_min_dot_css_.go \
 	bootstrap_dot_min_dot_js_.go \
 	jsontree_dot_js_.go \
+	jquery_dot_min_dot_js_.go \
```
edmc-ss, line 6: Not quite sure why you didn't preserve the order followed in .gitignore (or visa versa), but whatever... As you probably noticed, the entries are alphabetic so I guess this one is the correct ordering :-).

edmc-ss (approved):
I'd just fix the out-of-order in .gitignore... the rest looks great!

httpserver/Makefile
gerardgine: Oh, sorry, I didn't notice any specific order...

PR #498, opened 2020-05-26
Close connection to down rev clients

bschatz-swift:
If the client is not sending it's unique ID, sleep 1 second to slow down
the client from retrying and then close the connection.

Also, log a message with the source IP address of the client which is
misbehaving.


edmc-ss (approved):
Perfect!

PR #499, opened 2020-06-02
Remove swift-specs setup

tipabu:
That repo has been dead since 2016 or so.


gerardgine:
@tipabu I will try to bring up Runway with your changes later today. I don't think your changes will have any impact, but I guess it's worth checking. If everything works just fine, I'll merge.

PR #504, opened 2020-06-19
Fix memory leak and more unit tests

bschatz-swift:
* don't delete clientInfo from server.perClientInfo unless the completed queue is empty AND the socket has been closed.   If the socket is not close then we continue to queue completed tasks but we do not do trimming
* write unit tests to test short and TTL trimmers
* when trimming, don't push items on second list - just delete inline to avoid copies
* add some basic statistics - will be moved to bucketstats in next commit

### DIFF
--- a/retryrpc/server.go
+++ b/retryrpc/server.go
@@ -502,11 +502,12 @@ func (server *Server) trimTLLBased(ci *clientInfo, t time.Time) (numItems int) {
 		}
 	}
 	s := ci.stats
-	logger.Infof("ID: %v completedRequest len: %v completedRequestLRU len: %v previousHighestReplySeen: %v highestReplySeen: %v ",
-		"largestReplySize: %v largestReplySizeMethod: %v numRPCcompleted: %v numRPCretried: %v numRPCattempted: %v ",
+	logger.Infof("ID: %v completedRequest len: %v completedRequestLRU len: %v previousHighestReplySeen: %v highestReplySeen: %v "+
+		"largestReplySize: %v largestReplySizeMethod: %v numRPCcompleted: %v numRPCretried: %v numRPCattempted: %v "+
 		"numRPCinprocess: %v longest RPC: %v longest RPC Method: %v cntAddCompleted: %v cntRmCompleted: %v",
-		ci.myUniqueID, len(ci.completedRequest), ci.completedRequestLRU.Len(), s.largestReplySizeMethod, s.numRPCcompleted,
-		s.numRPCretried, s.numRPCattempted, s.numRPCinprocess, s.longestRPC, s.longestRPCMethod, s.cntAddCompleted, s.cntRmCompleted)
+		ci.myUniqueID, len(ci.completedRequest), ci.completedRequestLRU.Len(), ci.previousHighestReplySeen, ci.highestReplySeen,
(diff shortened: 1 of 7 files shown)


retryrpc/server.go
```diff
@@ -405,13 +433,16 @@ func (server *Server) trimCompleted(t time.Time, long bool) {
...
 		for e := l.Front(); e != nil; e = e.Next() {
 			key := e.Value.(string)
-			v := server.perClientInfo[key]
+			ci := server.perClientInfo[key]
 
-			v.Lock()
-			if v.isEmpty() {
+			ci.Lock()
+			ci.cCtx.Lock()
+			if ci.isEmpty() && ci.cCtx.serviceClientExited == true {
```
edmc-ss, line 440: Yummy! Love these one line changes for some challenging issues!!!

bschatz-swift: yes - very isolated
I have not added bucketstats yet. I am trying to do that today but wanted to get this merged

retryrpc/server.go
```diff
@@ -452,32 +484,31 @@ func (server *Server) trimAClientBasedACK(uniqueID string, ci *clientInfo) (numI
...
+			e = e.Next()
+			_ = ci.completedRequestLRU.Remove(eTmp)
+			numItems++
 		} else {
 			// Oldest is in front so just break
 			break
 		}
 	}
+	s := ci.stats
+	logger.Infof("ID: %v completedRequest len: %v completedRequestLRU len: %v previousHighestReplySeen: %v highestReplySeen: %v "+
```
edmc-ss, line 505: I see what you did here... You have your own "stats" struct that you rapidly update as operations occur... then, here, you "log" the results.

Can I say that I absolutely HATE logging stats? That is something a tool outside ProxyFS et. al. should do by scraping /stats and/or /metrics imho. Just my $0.02.

But what happened to using bucketstats (let alone stats)?

bschatz-swift: I am removing this logging it my next commit.   I wanted to get this merged first

edmc-ss (approved):
I'm good with all you did. My only comment is that I hope we can move away from the practice of periodically logging stats... and, instead, adopt the practice of using e.g. package stats... Prometheus is already able to scrape updates to package stats and provide a far more useful way to look at historical stats than we will ever find by simply logging stats periodically. Just my $0.02

bschatz-swift:
Are stats and bucketstats different?

I want to get rid of this log message ASAP.

edmc-ss:
**package stat** supports  simple operation counters with very rudimentary "bucketizing" if you'd like... It sends those stats to a statsd via UDP periodically/smoothly. It also is queryable via the **embedded HTTP server at "/metrics"** with the names of the counters converted from statsd convention (basically words are separated by periods) to Prometheus convention (basically words are separated by underscores).

**package bucketstats** is a more streamlined means to accumulate stats by bucketizing insertions. This is really good for logging things like I/Os or time durations since it will be able to report the distribution of stat updates very nicely. While there is no (as of today) means for Prometheus (or equivalent) to scrape the bucketized statistics, you can fetch them via the **embedded HTTP server at "/stats"**. You will see the buckets reported as Total-of-all-updates, Average-per-update, and a count of updates in each power-of-2 (I think always, but there may be other bases supported) bucket.

PR #507, opened 2020-06-26
Remove unneeded references to "KEEPALIVE" in retryrpc and its clients

edmc-ss:


edmc-ss:
Retracting this PR as it actually ends up with TCP/TLS using "no" keep alive (the default)

PR #509, opened 2020-07-01
Initial changes for server side Lease Management

edmc-ss:
Server side APIs are now defined... but the implementation is incomplete

Avoid infinite (uninterruptible) retry loops in retryrpc contacting ProxyFS...also allows prompt exiting

### DIFF
--- a/retryrpc/client.go
+++ b/retryrpc/client.go
@@ -13,6 +13,12 @@ import (
 	"github.com/swiftstack/ProxyFS/logger"
 )
 
+const (
+	ConnectionRetryDelayMultiplier = 2
+	ConnectionRetryInitialDelay    = 100 * time.Millisecond
+	ConnectionRetryLimit           = 8
+)
+
 // TODO - what if RPC was completed on Server1 and before response,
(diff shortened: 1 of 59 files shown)


retryrpc/client.go
```diff
@@ -391,6 +423,7 @@ func (client *Client) dial() (err error) {
 		client.connection.tlsConn.Close()
 		client.connection.tlsConn = nil
 	}
+
 	client.connection.tlsConn = tlsConn
 	client.connection.state = CONNECTED
 	client.connection.genNum++
```
bschatz-swift, line 431: The 3 lines setting the state to CONNECTED should move moved after the check for error from *sendMyInfo()*

If *sendMyInfo()* fails, we should NOT change the connection state since it could be in RETRANSMITTING and would open a subtle race condition where another goroutine thinks it needs to initiate retransmitting in parallel with this goroutine.

I think there is a bug in the existing code which we have not hit yet and I think by moving the setting of CONNECTED to after the call to sendMyInfo() we avoid the problem.

bschatz-swift (changes requested):

PR #511, opened 2020-07-03
Feature provisioning

charmster:
Changes to confgen to work with the provisioning API and accept volume groups without a VIP while still requiring a VIP for NFS and SMB volume groups.


confgen/api_internal.go
```diff
@@ -181,7 +181,7 @@ func computeInitial(envMap EnvMap, confFilePath string, confOverrides []string,
 
 	// Fetch pertinent data from Initial Config
 
-	_, localVolumeGroupMap, localVolumeMap, _, _, err = fetchVolumeInfo(initialConfMap)
+	_, localSMBVolumeGroupMap, localVolumeMap, _, _, err = fetchVolumeInfo(initialConfMap)
```
edmc-ss, line 184: Is it the case that:
1) A VolumeGroup may only contain SMB or only contain NFS or only contain non-SMB/non-NFS Volumes?
2) Why must we assume fetchVolumeInfo() returns only an SMB Volume Group Map - what about NFS/non-SMB/NFS?

confgen/api_internal.go
```diff
@@ -954,10 +982,18 @@ func fetchVolumeInfo(confMap conf.ConfMap) (whoAmI string, localVolumeGroupMap v
 
 	for volumeGroupName, volumeGroup = range globalVolumeGroupMap {
 		if whoAmI == volumeGroup.PrimaryPeer {
-			localVolumeGroupMap[volumeGroupName] = volumeGroup
 			for volumeName, volume = range volumeGroup.VolumeMap {
 				localVolumeMap[volumeName] = volume
 			}
+
+			var shared bool
```
edmc-ss: I don't think declaring a local scope var shared here follows the pattern in the codebase

confgen/api_internal.go
```diff
@@ -954,10 +982,18 @@ func fetchVolumeInfo(confMap conf.ConfMap) (whoAmI string, localVolumeGroupMap v
...
+			shared, err = IsVolumeGroupSharedSMB(confMap, volumeGroupName)
```
edmc-ss: I take it this discerning VGs for SMB vs non-SMB (and non-either) is because each SMB VG has it's own smb.conf (and VIP).

Is there a corresponding func to fetch the non-SMB VGs (presumably for populating the "connon" /etc/exports)?

confgen/api_internal.go
```diff
@@ -966,3 +1002,85 @@ func fetchVolumeInfo(confMap conf.ConfMap) (whoAmI string, localVolumeGroupMap v
...
+	return
+
+}
+
+// IsVolumeGroupSharedSMB Returns true if any volume in the volume group is
+// shared using SMB.  While we don't support having a volume be shared by both
+// SMB and NFS, it seems like two different volumes in a volume group could be,
+// so a volume group might be "shared" using both protocols.
+//
+func IsVolumeGroupSharedSMB(confMap conf.ConfMap, vgName string) (shared bool, err error) {
```
edmc-ss: I see here what is going on... but why is this called "Shared"? I get that the NFS form is...

confgen/sample-proxyfs-configuration/proxyfs-shares.conf
```diff
@@ -7,12 +7,12 @@ ReadCacheLineSize: 1048576
 ReadCacheWeight:   100
 PrimaryPeer: c67631b5-cb88-11e9-99da-0248604d6797
 VolumeList: vol-vg32-2
-VirtualIPAddr:  32.32.32.32
+VirtualIPAddr:  32.32.32.32/24
```
edmc-ss, line 10: Should we perhaps not name this "VirtualIPAddr" given that it also includes the netmask (in this case /24)?

headhunter/checkpoint.go
```diff
@@ -546,6 +549,8 @@ func (volume *volumeStruct) getEtcdCheckpointHeader() (checkpointHeader *Checkpo
 
 	checkpointHeaderEtcdRevision = getResponse.Kvs[0].ModRevision
 
+	logger.Infof("getEtcdCheckpointHeader: returning for volume '%s' etcdkey '%s' header %+v",
```
edmc-ss, line 552: Love these logging messages. I have half a mind to mirror them on the HEAD fetches from Swift.. but I hope we can eventually dispense with the .__checkpoint__ HEAD requests once we are assured all checkpoints are recorded in etcd.

edmc-ss (approved):
I really just had some questions. Looks great!

confgen/api_internal.go
bschatz-swift: Yes - I have the same questions

It seems like this function should be reorganized to take advantage of **IsVolumeGroupSharedSMB()** and **IsVolumeGroupSharedNFS()**

Only setup SMB and NFS if you have volume groups which use them.

confgen/api_internal.go
```diff
@@ -486,6 +486,7 @@ func computeVolumeSetChange(oldConfMap conf.ConfMap, newConfMap conf.ConfMap) (t
 	_, _, newLocalVolumeMap, _, _, err = fetchVolumeInfo(newConfMap)
 	if nil != err {
 		err = fmt.Errorf("In newConfMap: %v", err)
+		return
```
bschatz-swift, line 489: Does this work if there are no volumes and no VolumeGroups in the configuration?

confgen/api_internal.go
```diff
@@ -775,25 +782,46 @@ func populateVolumeGroup(confMap conf.ConfMap, globalVolumeGroupMap volumeGroupM
 			return
 		}
```
bschatz-swift, line 785: It seems like this function should be reorganized to take advantage of IsVolumeGroupSharedSMB() and IsVolumeGroupSharedNFS()

Perhaps it would simplify the function

confgen/api_internal.go
bschatz-swift: There is **IsVolumeGroupSharedNFS()** to tell you if the VG needs NFS.

confgen/api_internal.go
bschatz-swift: Are you saying these should be renamed **IsVolumeGroupSMB()** and **IsVolumeGroupNFS()**?

confgen/sample-proxyfs-configuration/proxyfs-shares.conf
```diff
@@ -7,12 +7,12 @@ ReadCacheLineSize: 1048576
 ReadCacheWeight:   100
```
bschatz-swift, line 7: I noticed that there is a file **pfs.txt** in confgen/sample-proxyfs-configuration.    I think we should remove that file since it appears to be out of date.

confgen/api_internal.go
charmster: Currently, if a volume group has NFS volumes and/or SMB volumes then it must have a Virtual IP.

The converse is not true.  For example, it is OK for an empty volume group to have a Virtual IP.  Originally, I did add a check that a volume group without NFS or SMB volumes must not have an VIP, but I took that out when I ran into a problem with an empty VG.  Since VG will be provisioned separately from volumes, it is always possible to have such an empty VG for a short period of time.  Or if a volume gets deleted i don't want config pushes to suddenly blow up.

the answer to the second question is "expediency".  the code doesn't need to know the list of volume groups with NFS volumes, or the list of volume groups without NFS or SMB volumes because it doesn't have to do anything for them, so why bother returning them?

Bob raises an interesting point!

localSMBVolumeGroupMap is used here:
```
        // Create per VG smb.conf files                                                                                  
        err = createSMBConf(initialDirPath, localSMBVolumeGroupMap)
        if nil != err {
                // TODO - logging                                                                                        
                return
        }
```

i think i felt `createSMBConf()` should be passed only the list of SMB volume containing volume groups.  but I could, instead, have modified the for loop inside to call `IsVolumeGroupSharedSMB()`:

```
        for _, volumeGroup = range localVolumeGroupMap {
                vipDirPath = initialDirPath + "/" + vipsDirName + "/" + volumeGroup.VirtualIPAddr.String()

                err = os.Mkdir(vipDirPath, confDirPerm)
                if nil != err {
                        return
                }

                fileName := vipDirPath + "/smb-VG-" + volumeGroup.VolumeGroupName + ".conf"
                f, openErr := os.OpenFile(fileName, os.O_APPEND|os.O_CREATE|os.O_WRONLY, smbConfPerm)
                if nil != openErr {
                        err = openErr
                        return
                }

                err = globalTplate.Execute(f, volumeGroup)
                if nil != err {
                        return
                }

                err = sharesTplate.Execute(f, volumeGroup)
                if nil != err {
                        return
                }
        }
```

(the original problem was the `mkdir` was failing when `volumeGroup.VirtualIPAddr.String()` returned "nil".  the second attempt to create the directory "nil" failed with EEXIST.)

so i could move the `IsVolumeGroupSharedSMB()` out of `fetchVolumeInfo()` and into `createSMBConf()`.  would that make you happier?

bschatz-swift: Generally - I think you should just make a call as to what the correct thing to do.

As a reviewer, I try to look at things from a different angle to see I can find bugs are simplify the code.

I could be entirely wrong - I just try to raise probing questions.

I always appreciate when someone looks at my code and says - “You know if you do this you may be able to simplify the code and fix this potential bug.”

Long story short - "Do the right thing!".

edmc-ss: I get our point about not needing to know the NFS Volumes in this case... since, as we've discussed, the /etc/exports has to have *ALL* served NFS Volumes in it whereas this function is dealing specifically with what goes in this VIPs/VG's smb.conf, right?

confgen/api_internal.go
charmster: It is correct that this doesn't follow the pattern in the codebase.

Whether its a good pattern is a different question that i won't debate here.  i'll change this.

confgen/api_internal.go
charmster: Ed is correct about the reason for checking whether a VG has any SMB shares.  If it does, then it needs per VIP smb.conf files.

Volumes that are shared via NFS go in /etc/exports, not volume groups.  that code walks the volume list and checks `len(volume.nfsClientList) > 0`.

confgen/api_internal.go
charmster: I don't know.  Bob's `IsVolumeSMB()` seems strange to me ... i don't think of a volume group being "SMB".

i could use the more verbose `IsVolumeGroupSharedViaSMB()`/`IsVolumeGroupSharedViaNFS()` i suppose ...

i'll make that change.

confgen/sample-proxyfs-configuration/proxyfs-shares.conf
charmster: We could call it `VirtualIPAddrInCIDRNotation` or, perhaps, `VirtualCIDRIPAddr`.

But that requries changes to the controller, monitord, and confgen and is really outside the scope of this work.

headhunter/checkpoint.go
charmster: Sounds good.

confgen/api_internal.go
charmster: Ed, you correct.

I rewrote the code so that `FetchVolumeInfo()` returns all volume groups and the `CreateSMBConf()` function then calls `IsVolumeGroupSharedSMB()`, but that had the undesirable consequence of requiring that I also pass `InitialConfMap` (the configuration map) when it was required before which meant the `templates.go` file where CreateSMBConf() lives then needed to import `github.com/swiftstack/ProxyFS/conf` whereas it didn't before.  I felt that was worse then the change making `FetchVolInfo()` less general.  I'll post the two diffs in Slack so you can compare them.

Perhaps the "right" way to fix this is to change the `VolumeGroup` struct to include booleans for NFS and SMB sharing, and then fill those in.  But that has its downsides as well.

confgen/api_internal.go
charmster: Hmmm.  By my reading of the code, lack of volumes and volume groups will not cause `fetchVolumeInfo()` to return an error.  If it did, we would have run into the problem with the same call 6 lines earlier.

Again by my reading of the code, it looks like `computeVolumeSetChange()` works fine if the `newLocalVolumeMap` or `oldLocalVolumeMap` is empty.

We're not using `computeVolumeSetChange()` anyway so I'm not going to spend time testing it.  I just added a return that seemed to be missing, though looking at the code we would wind up returning from the bottom of the function with `err != nil` so this is not actually a functional change.

charmster: In reference to "It seems like this function should be reorganized to take advantage of IsVolumeGroupSharedSMB() and IsVolumeGroupSharedNFS()":

I'm not sure that that would improve readability or maintainability.  I'd have to add a loop over volume groups and then construct an SMB volume group list.

confgen/api_internal.go
charmster: I could certainly add `IsVolumeGroupSharedViaSMB` and `IsVolumeGroupSharedViaNFS` bools to the `VolumeGroup` struct and handle it that way.

That would _probably_ be an improvement.  It adds some "distributed state", where the flag gets set in one place and checked in another, compared to "get the list of SMB volume groups and pass those to `createSMBConf()`", which may be clearer.  I could go either way.

Truthfully, I was just trying to get `etcd-mgmt` working with the minimal amount of changes to `confgen`.  Sorting all the issues between `monitord` and `confgen` took quite awhile ...

confgen/sample-proxyfs-configuration/proxyfs-shares.conf
charmster: Done.

PR #517, opened 2020-07-10
Add in support for the other /debug/pprof URLs into ProxyFS's & PFSAg…

edmc-ss:
…ent's embedded HTTP Servers

### DIFF
--- a/httpserver/request_handler.go
+++ b/httpserver/request_handler.go
@@ -157,8 +157,6 @@ func doGet(responseWriter http.ResponseWriter, request *http.Request) {
 		doGetOfMetrics(responseWriter, request)
 	case "/stats" == path:
 		doGetOfStats(responseWriter, request)
-	case "/debug/pprof" == path:
-		pprof.Index(responseWriter, request)
 	case "/debug/pprof/cmdline" == path:
 		pprof.Cmdline(responseWriter, request)
 	case "/debug/pprof/profile" == path:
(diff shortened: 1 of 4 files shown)


edmc-ss:
I believe:

1) Package pkg/profile appears to be superceded by std package pprof... so I think we can delete it
2) I've added support for the missing /debug/pprof targets in both ProxyFS's and PFSAgent's embedded HTTP Servers which probably means we can delete the optional "port 6060 (by default if not disabled or overridden)" HTTP Server in proxyfsd/daemon.go

In either case, let's agree on zero, one or both of these additional changes to go into this PR before merging it.

httpserver/request_handler.go
```diff
@@ -157,8 +157,6 @@ func doGet(responseWriter http.ResponseWriter, request *http.Request) {
 		doGetOfMetrics(responseWriter, request)
 	case "/stats" == path:
 		doGetOfStats(responseWriter, request)
-	case "/debug/pprof" == path:
-		pprof.Index(responseWriter, request)
 	case "/debug/pprof/cmdline" == path:
 		pprof.Cmdline(responseWriter, request)
 	case "/debug/pprof/profile" == path:
```
charmster, line 162: With the addition of the "HasPrefix()" case you can get rid of all of the other /debug/pprof* lines.

Same for the code in pfsagent.

charmster (approved):
This seems fine.  Note the one comment on the code I added.

As for the other questions you posed, I'd be happy to see the janky `github.com/pkg/profile` disappear.  I don't think I've ever successfully used it.

I left the "debug server" running on port 6060 around because it costs basically nothing.  I didn't think it was worth the effort to try and track the URL's that it provides in our code because they have changed over time.  There are actually a lot more of them then are listed on the documentation: allocs, block, cmdline, goroutine, heap, mutex, profile, threadcreate, and trace.

It seems like your "HasPrefix()" trick has solved that problem!

I'd prefer to keep the debug server running on port 6060 (monitord has the same one running on port 6059) because I'm used to it.  But if it really bothers you, you can delete it and I can adapt.

Note that I discovered that:
```
curl -i http://localhost:6060/debug/pprof/?
```
shows a list of possible ones.

httpserver/request_handler.go
edmc-ss: You would think that were true, huh?

I tried that and those targets are apparently not parsed for in the pprof.Index() function... very odd but true.

PR #520, opened 2020-07-25
Bugfix conf memorybloat

charmster:
Memory bloat due to fragmentation is bad.  Death to memory bloat!

Also, improve startup and log rotate log messaging.

### DIFF
--- a/conf/api.go
+++ b/conf/api.go
@@ -346,7 +346,7 @@ func (confMap ConfMap) UpdateFromFile(confFilePath string) (err error) {
 // To enable efficient comparisons, the elements of the ConfMap will be
 // sorted in the output (both by sectionName and by optionName).
 //
-func (confMap ConfMap) Dump() (buf string) {
+func (confMap ConfMap) Dump() (confMapString string) {
 	var (
 		confOption                ConfMapOption
 		confOptionName            string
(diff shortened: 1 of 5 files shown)


conf/api.go
```diff
@@ -360,7 +360,10 @@ func (confMap ConfMap) Dump() (buf string) {
 		confSectionNameSliceIndex int
 	)
 
-	buf = ""
+	// 1 Mibyte should be more than enough to hold the confMap so get the
+	// memory in 1 allocation
+	var buf strings.Builder
```
edmc-ss, line 365: As ever... when you modify an existing code block... it would be very helpful to follow the convention of the code being modified.

conf/api.go
```diff
@@ -394,20 +397,22 @@ func (confMap ConfMap) Dump() (buf string) {
...
+					buf.WriteString(", " + confOptionValue)
 				}
 			}
 
-			buf += "\n"
+			buf.WriteString("\n")
 		}
 	}
 
+	confMapString = buf.String()
```
edmc-ss, line 415: Since you inexplicably renamed the return argument, what does this line accomplish?

edmc-ss (commented):
Not sure what problem this solves... is memory getting fragmented or something by repeated calls to Dump()?

conf/api.go
```diff
@@ -346,7 +346,7 @@ func (confMap ConfMap) UpdateFromFile(confFilePath string) (err error) {
 // To enable efficient comparisons, the elements of the ConfMap will be
 // sorted in the output (both by sectionName and by optionName).
 //
-func (confMap ConfMap) Dump() (buf string) {
+func (confMap ConfMap) Dump() (confMapString string) {
```
edmc-ss, line 349: I didn't fully grok this the first time... this change looks great!

conf/api.go
edmc-ss: Now I see it... your use of buf as a strings.Builder nicely avoids changes throughout the code... then you "unwrap" it to return the desired string to callers with this line

PR #538, opened 2020-09-30
Move eventlet.Timeout handling into JsonRpcClient.call()

tipabu:
We previously documented that `call()` would raise `RpcTimeout`s; let's make that actually be true.

This also gives us a handy place to centralize our error formatting. As a drive-by, make sure that we include the ip/port we were trying to talk to in the error.


edmc-ss:
Merged "locally" as part of "prs_539_and_538" (PR 540)

PR #539, opened 2020-09-30
Shuffle proxyfsd addresses before doing bimodal check

tipabu:
Previously, we would iterate through them in an arbitrary but deterministic order, which could lead to issues if we keep hitting timeouts trying to talk to what happens to be the first one.

### DIFF
--- a/pfs_middleware/pfs_middleware/bimodal_checker.py
+++ b/pfs_middleware/pfs_middleware/bimodal_checker.py
@@ -23,6 +23,7 @@ its own, but it is required for the other middlewares to work.
 """
 
 import eventlet
+import random
 import socket
 import time
 
@@ -51,7 +52,7 @@ class BimodalChecker(object):
                           in conf.get('proxyfsd_host', '127.0.0.1').split(',')]
         self.proxyfsd_port = int(conf.get('proxyfsd_port', '12345'))
 
-        self.proxyfsd_addrinfos = []
+        self.proxyfsd_addrinfos = set()
         for host in proxyfsd_hosts:
             try:
                 # If hostname resolution fails, we'll cause the proxy to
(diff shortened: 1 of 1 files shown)


dbishop:
So this PR addresses always hitting the same proxyfsd (which I guess is bad if that one hit is bad), but after this PR, we're still bad if we get unlucky and still hit a bad proxyfsd?  I guess there's a separate body of work to prevent a 503 in the wsgi stack if a chosen proxyfsd is bad (in some time-consuming sense)?

edmc-ss:
Yeah, the JIRA mentioned tries to point out the various concerns/areas-of-improvement...
The randomization not reusing the same seed each time would mean that retries may make a more fortunate selection...

Moving forward, if PFSAgent is the only entity using "BiModal" Accounts, we can ensure such retries are sufficient, but it would be preferable to be more tolerant (e.g. by increasing or eliminating the time-based aspect of pfs_middleware's bimodal_checker.py code (see RPC_FINDER_TIMEOUT_DEFAULT = 3.0 constant declaration at line 39 and the commensurate tunable in the [filter:pfs] section key `rpc_finder_timeout`. Also, as the list of ProxyFS-enabled nodes grows, knowing about and trying every single such node may exceed any reasonable timeout. We need a better mechanism for sure.

tipabu:
Yup -- the idea here is just to get a minimal fix that should allow a client that's willing to retry on 503s to eventually get a good response (as long as the volume they're trying to reach is still up and reachable from whatever proxy's handling the request). Note that [there's caching](https://github.com/swiftstack/ProxyFS/blob/e019e7d19403c092713a21799eb580008dab7d86/pfs_middleware/pfs_middleware/bimodal_checker.py#L138-L143) we do once we get a response, so I'd expect having this and a retrying client should make things fairly usable.

FWIW, most Swift and S3 clients will retry at least a few times on a 503 -- the goal here is just to make sure there's a decent chance of that retry succeeding.

edmc-ss:
Merged "locally" as part of "prs_539_and_538" (PR 540)

PR #558, opened 2020-11-30
Bugfix 10 k volumes

charmster:
Address issues in ProxyFS discovered when trying to serve 10,000 volumes in a cluster (more than ~1,600 on a single node).

Add a new config variable to enable the liveness checker, `cluster.LivenessCheckerEnabled`, and disable it by default until we address the panic in the code.

Add back the debug HTTP server so we can get goroutine stack tracebacks before `transitions.Up()` has finished, (which is taking ~15 min on my AWS cluster with ~1,500 volumes, while we are mounting filesystems).

Arm the signal handler so they can catch SIGHUP before transitions.Up() has completed.

Slightly unrelated: add `conf.SetOptionIfMissing()` to support monitord setting LivenessCheckerEnabled if the Controller hasn't and `conf.SetSectionIfMissing()`, to be used to set backing policies for NGC Batch volumes if the cluster hasn't.

### DIFF
--- a/CONFIGURING.md
+++ b/CONFIGURING.md
@@ -77,6 +77,7 @@ For each of the keys supported, the following table will list whether or not its
 |                                           | HeartBeatMissLimit                       | No           | 3                  | Yes                      | No                           |
 |                                           | MessageQueueDepthPerPeer                 | No           | 4                  | Yes                      | No                           |
 |                                           | MaxRequestDuration                       | No           | 1s                 | Yes                      | No                           |
+|                                           | LivenessCheckerEnabled                   | No           | false              | Yes                      | Yes                          |
 |                                           | LivenessCheckRedundancy                  | No           | 2                  | Yes                      | No                           |
 |                                           | LogLevel                                 | No           | <i>None</i>        | Yes                      | No                           |
 | Peer:<i>PeerName</i>                      | PublicIPAddr                             | Yes          |                    | Yes                      | Yes but WhoAmI should remain |
--- a/httpserver/setup_teardown_test.go
+++ b/httpserver/setup_teardown_test.go
@@ -56,6 +56,7 @@ func testSetup(t *testing.T) {
 		"Cluster.HeartBeatMissLimit=3",
 		"Cluster.MessageQueueDepthPerPeer=4",
 		"Cluster.MaxRequestDuration=1s",
+		"Cluster.LivenessCheckerEnable=true",
 		"Cluster.LivenessCheckRedundancy=2",
 		"FSGlobals.VolumeGroupList=",
 		"FSGlobals.CheckpointHeaderConsensusAttempts=5",
--- a/httpserver/setup_teardown_test.go
+++ b/httpserver/setup_teardown_test.go
@@ -90,6 +90,7 @@ func testSetup(t *testing.T) {
 		"RamSwiftInfo.MaxObjectNameLength=1024",
 		"RamSwiftInfo.AccountListingLimit=10000",
 		"RamSwiftInfo.ContainerListingLimit=10000",
+
 		"HTTPServer.TCPPort=53461",
 	}
 
--- a/proxyfsd/daemon.go
+++ b/proxyfsd/daemon.go
@@ -51,10 +51,8 @@ func Daemon(confFile string, confStrings []string, errChan chan error, wg *sync.
 	// available if transitions.Up() hangs (the embedded http server will
 	// return http.StatusServiceUnavailable (503) during the transition.
 	debugServerPortAsUint16, err := confMap.FetchOptionValueUint16("ProxyfsDebug", "DebugServerPort")
-	if nil != err {
-		debugServerPortAsUint16 = 6058 // TODO: Eventually set it to zero
-	}
-	if uint16(0) != debugServerPortAsUint16 {
+	if nil != err && debugServerPortAsUint16 != 0 {
+
 		debugServerPortAsString := fmt.Sprintf("%d", debugServerPortAsUint16)
 		logger.Infof("proxyfsd.Daemon() starting debug HTTP Server on localhost:%s", debugServerPortAsString)
 		go http.ListenAndServe("localhost:"+debugServerPortAsString, nil)
--- a/proxyfsd/daemon.go
+++ b/proxyfsd/daemon.go
@@ -60,6 +60,18 @@ func Daemon(confFile string, confStrings []string, errChan chan error, wg *sync.
 		go http.ListenAndServe("localhost:"+debugServerPortAsString, nil)
 	}
 
+	// Arm signal handler used to catch signals
+	//
+	// Note: signalChan must be buffered to avoid race with window between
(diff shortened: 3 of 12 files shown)


bschatz-swift (approved):

CONFIGURING.md
```diff
@@ -77,6 +77,7 @@ For each of the keys supported, the following table will list whether or not its
 |                                           | HeartBeatMissLimit                       | No           | 3                  | Yes                      | No                           |
 |                                           | MessageQueueDepthPerPeer                 | No           | 4                  | Yes                      | No                           |
 |                                           | MaxRequestDuration                       | No           | 1s                 | Yes                      | No                           |
+|                                           | LivenessCheckerEnabled                   | No           | false              | Yes                      | Yes                          |
```
edmc-ss, line 80: Awesome - thanks!

httpserver/setup_teardown_test.go
```diff
@@ -56,6 +56,7 @@ func testSetup(t *testing.T) {
 		"Cluster.HeartBeatMissLimit=3",
 		"Cluster.MessageQueueDepthPerPeer=4",
 		"Cluster.MaxRequestDuration=1s",
+		"Cluster.LivenessCheckerEnable=true",
```
edmc-ss, line 59: I'm wondering if we might want to always disable LivenessChecker in our test setup's...
I do not think any of the "unit tests" actually test this... so its fine either way...

proxyfsd/daemon.go
```diff
@@ -45,6 +46,32 @@ func Daemon(confFile string, confStrings []string, errChan chan error, wg *sync.
 		return
 	}
 
+	// Optionally launch an embedded HTTP Server for Golang runtime access;
```
edmc-ss, line 49: Seems to me that you wanted to be able to disable the Liveness Checker (I really don't know why), so you included a mechanism to do so.

I should think it *ATLEAST* as desirable to be able to disable this "optional" Debug Server port. Indeed, having said it was "optional", it appears it is not.

Let me propose this compromise:
1) This embedded HTTP Server be made strictly optional (e.g. make an enabled bool config var or, alternatively, if no port is specified, the behavior is to disable it), and
2) Disable this by default.

Of course the counter proposal would be to *NOT* provide this (in checked in code), and, thus, remove it from CONFIGURING.md :-).

But lets assume we want it in... then certainly we want to ensure that e.g. test setup code either explicitly or implicitly disables it.

Make sense?

edmc-ss (changes requested):
This looks great @charmster ... but, as you'll see, I did request a small compromise to this Debug HTTP Server implementation.

edmc-ss:
One more comment @charmster ... I see that you allowed for the disabling of Liveness Checker... really don't want that in there... but perhaps you have a solid reason? In any event, what happens if it is disabled and somebody "queries" the HTTP Server's access to that (e.g. via curl http://localhost:15346/liveness)?

httpserver/setup_teardown_test.go
charmster: I left it enabled to, hopefully, reduce "bit rot" so it gets at least little bit of exercise.

Though i'm not sure how useful that is in a "single node" cluster where it has nobody to talk to and none of the tests try to talk to it.  But at least the init/deinit code gets called.

Since you say it's fine either way, i'll leave this enabled.

proxyfsd/daemon.go
charmster: By my reading of the code, the debug HTTP server can be disabled by setting the port to 0:

        debugServerPortAsUint16, err := confMap.FetchOptionValueUint16("ProxyfsDebug", "DebugServerPort")
        if nil != err {
                debugServerPortAsUint16 = 6058 // TODO: Eventually set it to zero                                        
        }
        if uint16(0) != debugServerPortAsUint16 {
                debugServerPortAsString := fmt.Sprintf("%d", debugServerPortAsUint16)
                logger.Infof("proxyfsd.Daemon() starting debug HTTP Server on localhost:%s", debugServerPortAsString)
                go http.ListenAndServe("localhost:"+debugServerPortAsString, nil)
        }

if "ProxyfsDebug.DebugServerPort" is not specified, then  debugServerPortAsString gets set to "6058".  But if it is specified, and it's 0, then the debug server is not started -- so i would say that it's optional.

i can change it to disabled by default and have `monitord` enable it in the proxyfs config files that it generates.  but my inclination would be to explicitly enable it in the unit tests as well, since the more elaborate ones sometimes have bugs that cause them to hang.

do you feel strongly that it should be disabled in unit tests?

also, if we disable it by default then it will not be available (by default) in the SAIO environment unless i change the default config files to enable it.

i can do that.

Do you have strong feelings about disabling it for unit tests?  or for the SAIO environment?

httpserver/setup_teardown_test.go
charmster: i thought i had already mentioned why i disabled the liveness detector ... for some damn reason it causes a panic when the number of volumes served by a proxyfs process goes beyond ~1,500 volumes.  it looks like this:

```
panic: sendMsg() called for excessive len(msgBuf) == 361741 (must be <= 352410)
--
goroutine 1907977 [running]:
github.com/swiftstack/ProxyFS/liveness.doFollower()
/home/swift/code/ProxyFS/src/github.com/swiftstack/ProxyFS/liveness/states.go:373 +0x1202
github.com/swiftstack/ProxyFS/liveness.stateMachine()
/home/swift/code/ProxyFS/src/github.com/swiftstack/ProxyFS/liveness/states.go:16 +0x27
created by github.com/swiftstack/ProxyFS/liveness.(*globalsStruct).SignaledFinish
/home/swift/code/ProxyFS/src/github.com/swiftstack/ProxyFS/liveness/config.go:840 +0x12f5
", "stdout": ""
```

i did a little bit of digging before deciding to punt the issue (by disabling liveness checker) and move on.

i ran wireshark to collect all the packets sent on that port.  i was anticipating that they would actually grow in size until they eventually hit the limit of 352,410 and i wanted to see what was filling them up.  instead, they stubbornly stayed small as i added volumes(1,400 bytes?) until proxyfs paniced with the same error.

proxyfsd/daemon.go
charmster: OK, i discovered that it is already explicitly enabled in the SAIO environment, though at a different port number (i changed the default from 6060 to 6058 because 6060 is in all the go documentation and thus more likely to be in use already).  the comment was already there.

```
# DebugServerPort will be served by an embedded HTTP Server responding to requests                                       
# as described in package net/http/pprof documentation. If not defined, the embedded                                     
# HTTP Server is not launched.                                                                                           
#                                                                                                                        
[ProxyfsDebug]
ProfileType:     None
DebugServerPort: 6058
```

in the unit tests, daemon_test.go is the only unit test that can usefully enable the go debug server since it is the only place where `Daemon()` is invoked.  I added:
```
                "ProxyfsDebug.DebugServerPort=6058",
```
to that test.

PR #571, opened 2021-01-21
Re license

edmc-ss:


sait/bin/provision_middleware
```diff
@@ -1,4 +1,7 @@
 #!/bin/bash
+\
```
gerardgine: Is this `\` expected?

gerardgine (commented):
There's just 1 line in 1 file that looks suspicious to me (see comment below). There rest is a go.

Please note: I didn't check anything under `vendor` that doesn't come from SwiftStack.

sait/bin/provision_middleware
edmc-ss: Nice catch... no it was not.

Here's an example of a script file w/out a suffix but with a "#!/bin/bash" 1st line... I obviously did that one manually... and fumble fingered... and didn't test this particular script.

PR #572, opened 2021-01-25
Adding TLS support to proxyfsd

bschatz-swift:


etcdclient/api.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"log"
+	"os"
+	"time"
+
+	etcd "go.etcd.io/etcd/clientv3"
+	etcdtransport "go.etcd.io/etcd/pkg/transport"
+)
+
+const (
```
edmc-ss, line 12: These "cert_path" items should be passed in the .conf imho.

etcdclient/api.go
```diff
@@ -0,0 +1,49 @@
...
+	// Certs are named based on the host name.
+	h, _ := os.Hostname()
+	certFile := certPath + "node-" + h + ".pem"
+	keyFile := certPath + "node-" + h + "-key.pem"
+
+	tlsInfo := etcdtransport.TLSInfo{
+		CertFile:      certFile,
+		KeyFile:       keyFile,
+		TrustedCAFile: caFile,
+	}
```
charmster: If New() took a `*etcdtransport.TLSInfo` argument then having that value be nil could indicate a simple TCP connection is used.

but then the code that is highlighted here either 1) gets repeated in multiple locations or 2) gets moved to some other function, which somewhat negates the value of creating `etcdclient`.

it's your call what you want to do here, if anything ...

charmster (commented):
Seems fine.

It will be good when we can move the "real" `etcdclient` package to proxyfs.

etcdclient/api.go
bschatz-swift: I have made the changes for tlsConfig in a new push.

I will file a Jira to make the change to have "cert_path" be a .conf entry and fix it as my top priority as soon as this merges.  (I want to complete the story so that we can start testing)

PR #578, opened 2021-01-30
Add EtcdCertPath to conf file

bschatz-swift:


etcdclient/api.go
```diff
@@ -13,8 +13,7 @@ import (
 )
 
 const (
-	certPath      = "/etc/ssl/etcd/ssl/"
-	trustedCAFile = certPath + "ca.pem"
+	trustedCAFile = "/ca.pem"
```
edmc-ss, line 16: I believe you should not start the filename with a slash... just prepend it (i.e. append trustedCAFile to the path)

Indeed, it looks like that's what you do...so (by default) you are ending up with a path that looks like "/etc/ssl/etcd/ssl//ca.pem" (note the double slash before "ca")... probably not what you want...

edmc-ss (changes requested):
Just that one comment (the leading slash on the ca file name that shouldn't be there but just happens to be "ok")...

etcdclient/api.go
bschatz-swift: Normally, I would not want this.    

However, I was worried about someone coming along later and removing the trailing "/" from certPath.   This is defensive programming.

bschatz-swift: @edmc-ss Do you understand why I did that?

edmc-ss: I'm just saying you have this backwards... :-)
How about this... rename the field "certPath" to "certDir"... so it is explicit it is a dir
Then, if the user specifies the trailing slash, it wouldn't matter if, instead, you computed your path by:
   certDir + "/" + trustedCAFile
Filenames don't have slashes...

bschatz-swift: Okay - I changed all places from "certPath" to "certDir" so that it is clearer

Let me know what you think

edmc-ss: Ah... but you left the leading '/' on trustedCAFile's value :-)

It's just a pet peive of mine... I can live with it this way...

edmc-ss (approved):

PR #595, opened 2021-03-16
Convert check_pinned_packages & pfs-swift-load-plot to py3

gerardgine:


edmc-ss:
Is it safe to assume I need to match some of these changes in SAIO et. al.?

PR #601, opened 2021-03-30
bump swift to ss-release-2.27.0.1

clayg:


clayg:
so as it turns out we're not releasing 2.27.0.1 to PBSS - there's going to be a 2.27.0.2 on ~4/12 - since this hasn't merged; I think we should just close it!

PR #609, opened 2021-04-14
Automatically get latest Swift tag in SAIO/SAIT

gerardgine:
It is now ready for review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,7 +198,7 @@ pre-generate:
 
 python-test:
 	cd meta_middleware && tox -e lint
-	cd pfs_middleware && tox -e py27-release,py27-minver,py36-release,lint
+	cd pfs_middleware && LATEST_SWIFT_TAG=` git ls-remote --refs --tags https://github.com/NVIDIA/swift.git | cut -d '/' -f 3 | grep ^ss-release- | sort --version-sort | tail -n 1 ` tox -e py27-release,py27-minver,py36-release,lint
 
 test:
 	@set -e; \
--- a/saio/container/Dockerfile
+++ b/saio/container/Dockerfile
@@ -3,7 +3,7 @@
 
 FROM centos:7.4.1708
 
-ARG SwiftVersion=ss-release-2.27.0.2
+ARG SwiftVersion
 ARG GolangVersion=1.15.5
 ARG ProxyFS_Version=stable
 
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -198,7 +198,7 @@ pre-generate:
 
 python-test:
 	cd meta_middleware && tox -e lint
-	cd pfs_middleware && tox -e py27-release,py27-minver,py36-release,lint
+	cd pfs_middleware && LATEST_SWIFT_TAG=` git ls-remote --refs --tags https://github.com/NVIDIA/swift.git | cut -d '/' -f 3 | grep ^ss-release- | sort --version-sort | tail -n 1 ` tox -e py27-release,py27-minver,py36-release,lint
```
edmc-ss, line 201: Really cool that you were able to figure out the TOX-specific change, too!

saio/container/Dockerfile
```diff
@@ -111,7 +111,7 @@ RUN pip install -r test-requirements.txt
 WORKDIR /home/swift
 RUN git clone https://github.com/NVIDIA/swift.git
 WORKDIR /home/swift/swift
-RUN git checkout $SwiftVersion
+RUN if [[ -n "$SwiftVersion" ]] ; then git checkout $SwiftVersion ; else git checkout $( git describe --tags --abbrev=0 $(git rev-list --tags --max-count=1) ) ; fi
```
edmc-ss, line 114: I assume this works...

Key thing, I think, that such a derivation would necessarily affect the Image "TAG" that gets specified as the result of the "docker build" command... perhaps that would necessitate a different approach than this (i.e. the version/tag derivation would be done in a "wrapper" around "docker build" that would, instead, explicitly specify the SwiftVersion arg to the Dockerfile... and, of course, specify the target name of the built Docker Image.

saio/vagrant_provision.sh
```diff
@@ -228,7 +228,8 @@ echo "export ST_KEY=testing" >> ~vagrant/.bash_profile
 cd ~swift
 git clone https://github.com/NVIDIA/swift.git
 cd swift
-git checkout ss-release-2.27.0.2
+export SWIFT_TAG="$( git describe --tags --abbrev=0 $(git rev-list --tags --max-count=1) )"
+git checkout $SWIFT_TAG
```
edmc-ss, line 232: Ditto.

sait/vagrant_provision.sh
```diff
@@ -230,7 +230,8 @@ echo "export ST_KEY=testing" >> ~vagrant/.bash_profile
 cd ~swift
 git clone https://github.com/NVIDIA/swift.git
 cd swift
-git checkout ss-release-2.27.0.2
+export SWIFT_TAG="$( git describe --tags --abbrev=0 $(git rev-list --tags --max-count=1) )"
```
edmc-ss, line 233: Ditto.

edmc-ss (changes requested):
Basically, I think the derivation of the SwiftVersion should be done *OUTSIDE* of the Dockerfile/provitioning script... and, in so doing, affect the "tag" applied to the created Docker Image.

gerardgine:
As discussed in real life, all your points are perfectly valid, but they're also outside the scope of what we're trying to achieve in this PR.

Merging!

PR #629, opened 2021-07-01
Breakout loggers

bschatz-swift:


retryrpc/api.go
```diff
@@ -320,10 +332,9 @@ type ClientConfig struct {
 	Callbacks                interface{}   // Structure implementing ClientCallbacks
 	DeadlineIO               time.Duration // How long I/Os on sockets wait even if idle
 	KeepAlivePeriod          time.Duration // How frequently a KEEPALIVE is sent
+	Logger                   *log.Logger   // If nil, defaults to log.New()
```
edmc-ss, line 335: I take it this "eases" integration:
1) Unmodified clients will get the "zero value" nil for this field automatically and, thus,
2) Use the default log.New() logger

I guess this is a nice way to change things w/out having to fix the clients... but each of those clients should be fixed, right?

I'm not worried about them... but I'll certainly be changing iclient (and, similarly, imgr) to supply an appropriate Logger: value.

edmc-ss (commented):
I know you've merged Bob... but I was looking for the corresponding Logger field/arg on the server side of things... and didn't find it. Am I missing something?

bschatz-swift:
Looks like I missed that :(

I know that when I was looking at the unit tests, it appeared to be a bunch of work to get the unit tests to pass the appropriate logger.   Therefore, I did not do it there because I was worried about spending a bunch of time on code which may be retired.

However, I should have done it in the other clients OR at least mentioned it.

